PR #61, opened 2015-05-10
Accessory factories, local variables for characteristics and callback…

snowdd1:
… on instanciating characteristics

This commit introduces a _new type of accessory files_, called accessory factory files (_accfactory.js). Those return an Array of accessory instead of a single accessory.
Additionally, accessory files can return local variables in an locals object, and can define an onRegister callback to register their characteristics object with some Event handler.

BridgedCore.js
--------------
inserted section to handle accessory factory files in accessories
folder. Accessory factories return an array of accessory objects instead
a single accessory object.
inserted onRegister optional callback property for accessories. This
callback is invoked in Characteristics.js in the constructor, and is
used to register the accessory with other modules, to be notified when
the state changes. With that Siri is able to answer questions like "Is
the light in the kitchen still switched on?", even when the kitchen
light state was changed externally.
inserted local optional property object for passing non-standard
parameters to the characteristics object.

Characteristics.js
------------------
inserted onRegister optional callback (see BridgedCore comment above)
inserted local optional property object (see BridgedCore comment above)

Core.js
-------
see BridgedCore.js

### DIFF
--- a/BridgedCore.js
+++ b/BridgedCore.js
@@ -12,7 +12,7 @@ var targetPort = 51826;
 var fs = require('fs');
 var path = require('path');
 
-var accessoriesJSON = []
+var accessoriesJSON = [];
 
 accessoriesJSON.push(require("./accessories/Bridge_Info.js").accessory);
 // Get user defined accessories from the accessories folder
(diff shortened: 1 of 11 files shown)


KhaosT:
Thanks :)

snowdd1:
Hello, I've seen you merged both my commits. The second contains KNX specific extensions which need additional dependencies. Do we want to have this in the master branch?
KNX is an association of (mostly) European hardware manufacturers with a common (wired) bus protocol.
REVERT: I've just seen you already put a .sample. in the filename. Thanks.

alistairg:
I'm trying to work out how to use onRegister - but I get "NotEventEnabled" when I call updateValue() on the callback. What am I doing wrong?

I can't see how to get the "eventEnabled" flag correctly set.... it doesn't seem to be related to "supportsEvents"?

snowdd1:
Have a look at (closed) issue #63. 
The app has to advertise its event capability as well. MyTouchHome does not, for example; though it works if you ignore the eventenabled setting node-side.

PR #82, opened 2015-06-25
Async Characteristic Reads

maddox:
This makes the characteristic reads callback based. This way we can have our `onRead` functions be async. We're fetching state here, so its useful to be able to use callbacks.

Here's an example in use:

```js
onRead: function(callback) {
  this.getBrightness(function(level){
    callback(level);
  });
}
```

## Notes

Ok, I barely know what I'm doing, but I basically just went up the stack making sure everything supported using a callback. The only place I got in trouble was in `Server.js` where the `JSON` is written to the response. The callback style works fine, the only problem is, on the first call through this stack, the `onRead` hasn't been defined yet. So everything just goes to 💩 and hangs. 

I'm not sure if this is actually a bug or what. I just handled it by sending back an empty string of `JSON`, if there `onRead` us undefined. This handles the problem, and on the next call through this stack, things work just fine. ¯\(°_o)/¯

### DIFF
--- a/AccessoryController.js
+++ b/AccessoryController.js
@@ -28,6 +28,10 @@ AccessoryController.prototype = {
 		}
 		return accessory;
 	},
+	isReadyForJson: function isReadyForJson(iid){
+		var charObject = this.objects[iid];
+		return charObject.onRead ? true : false;
+	},
 	jsonPresentation: function jsonPresentation() {
 		var servicesObjects = [];
 		for (var i = 0; i < this.services.length; i++) {
@@ -43,19 +47,20 @@ AccessoryController.prototype = {
 		}
 		return JSON.stringify(dict);
 	},
-	jsonForCharacteristicUpdate: function jsonForCharacteristicUpdate(aid, iid) {
+	jsonForCharacteristicUpdate: function jsonForCharacteristicUpdate(aid, iid, callback) {
 		var charObject = this.objects[iid];
-		var charValue = charObject.valueForUpdate();
-		var respDict = {
-			characteristics: [
-				{
-					aid: aid,
-					iid: iid,
-					value: charValue
-				}
-			]
-		}
-		return JSON.stringify(respDict);
+		var charValue = charObject.valueForUpdate(function(value){
+			var respDict = {
+				characteristics: [
+					{
+						aid: aid,
+						iid: iid,
+						value: value
+					}
+				]
+			}
+			callback(JSON.stringify(respDict));
+		});
 	},
 	processSingleCharacteristicsValueWrite: function processSingleCharacteristicsValueWrite(update, peer) {
 		var update_char = update;
--- a/Characteristic.js
+++ b/Characteristic.js
@@ -128,17 +128,19 @@ Characteristic.prototype = {
 			}
 		} else { console.log("Characteristics.js:updateValue():NotEventEnabled"); }
 	},
-	valueForUpdate: function valueForUpdate() { // reading values FROM THE DEVICE, better: from this object
+	valueForUpdate: function valueForUpdate(callback) { // reading values FROM THE DEVICE, better: from this object
 		console.log("Characteristics.js:valueForUpdate(): called, Siri has asked for the accessory's status");
+
 		if (this.onRead) {
 			console.log("Characteristics.js:valueForUpdate(): invoking callback");
-			var temp = this.onRead();
-			this.value = temp ? temp : this.value;
+			this.onRead(function(value){
+				this.value = value;
+				console.log("Characteristics.js:valueForUpdate(): called, Siri has asked for the accessory's status: returning " + this.value);
+				callback(this.value);
+			});
 		}
-		console.log("Characteristics.js:valueForUpdate(): called, Siri has asked for the accessory's status: returning " + this.value);
-		return this.value;
 		//
-		
+
 	}
 };
 
--- a/Server.js
+++ b/Server.js
@@ -106,8 +106,21 @@ HAPServer.prototype = {
 			var iids = requestDetails.query.substring(3).split(".");
 			var accessoryId = parseInt(iids[0]);
 			var characteristicId = parseInt(iids[1]);
-			response.write(this.accessoryController.jsonForCharacteristicUpdate(accessoryId,characteristicId));
-			response.end();
+
+			// For some reason, this is called before an accessory has had time to actually
+			// hook up it's callbacks. So we check first to see if it's ready.
+			// If it is, we attempt the query. If it isn't, we just return empty JSON.
+			// ¯\(°_o)/¯ It doesn't seem to hurt anything.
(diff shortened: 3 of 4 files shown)


maddox:
:dancer:

KhaosT:
@maddox Would you mind to add a case to support accessories that doesn't provide an onRead handler? Due to my current IP Agreement I cannot make any commit with actual code modification ... Thanks :)

maddox:
Oh, that's tricky.

```
	isReadyForJson: function isReadyForJson(iid){
		var charObject = this.objects[iid];
		return charObject.onRead ? true : false;
	}

``` 

This was made specifically to get around the weirdness I mentioned in the pull. But yeah, I guess for accessories that NEVER have this defined, it'll get into trouble.

Do you have any idea why the first time this is called, the `onRead` and `onUpdate` aren't wired up yet? I'm not familiar with the internals enough to really figure it out. I stared at it for a long time, heh.

maddox:
is `jsonForCharacteristicUpdate` called for the simple characteristics like `name` and `manufacturer`?

maddox:
Hmm, is this not good enought - https://github.com/KhaosT/HAP-NodeJS/commit/f8064528298345057154cb48a3072c79a86606e5#diff-85a092cec8485361f30d8bda6dcc4c71L133?

snowdd1:
How long is Siri going to wait for an answer? I tried the onRead with the KNX bus system, but when asking Siri for the state of multiple devices ( "Siri which lights are switched on upstairs?") the answer from the bus takes multiple seconds (send read request, wait for answer, for each device). Siri almost always gave up and told me my lights didn't answer.
But maybe I did something else wrong?
I navigated around this by the monitor/onRegister design to _immediatly_ update the value in the characteristics whenever a state change happens on the KNX bus. That way event enabled apps can show the state change without actively polling the status,too. This is probably a way for other systems which deal with a larger quantity of devices (I have about 40 light switches on the bus).
Please consider this before putting to much work in the "callback hell" as @maddox wrote in @nafarina 's issues.

nfarina:
Any reason not to allow both sync and async depending on accessory implementation?

nfarina:
For instance, the callback parameter could be executed immediately from within `onRead` and the calling function (HAP-NodeJS library) could be set up to detect this situation and return immediately.

snowdd1:
Possible. But if you want to have an easy-to-configure template you cannot rely on end users to distinguish between "I have only a few devices" and "I have to many devices". And as I wrote: Don't you want to have your homekit app have the status shown up-to-date?

In the characteristics.js there is a private member "value" and the methods for external  updates - you just have to call them if and when an update event occurs. Of course, you have to have some event hook to put your handlers on (And here my java script knowledge thins out - I did use an already set up framework for the KNX bus system). But I assume it is not to complicated to create event routines also for other systems.
The onRead callback I really programmed just to show it's possible, and I still do not encourage its use in a more-then-one-device environment.

snowdd1:
@maddox I had a look at the way you used the onRead callback, and I had knot in my brain. 
Are you sure you don't run into hundreds of race conditions when using async callbacks within the https answer creation? To me it looks overly complicated (but of course because I didn't write it..!)

nfarina:
I think it's great if the the values can always be "pushed" in realtime and returned synchronously, and this may be easy for some systems, but lots of existing products that weren't designed with HomeKit in mind will have trouble. For instance, fetching the status of a Lockitron lock or Xfinity security system involves a round-trip to a cloud server - I wouldn't want to run a polling operation every few seconds just to keep state updated proactively; this may run into API limits and use unnecessary bandwidth. And these are products where you won't have "lots of them" either. It would be really great if both options could be supported!

snowdd1:
Ok, you cannot get a push notification on status changes?

snowdd1:
@nfarina What about having this discussion in your own issues, for example #6 ?

nfarina:
Sure that's fine. And to be fair, I'd like to look through the existing accessories supported by homebridge to see how many of them would be able to do synchronous onRead.

maddox:
> I think it's great if the the values can always be "pushed" in realtime and returned synchronously, and this may be easy for some systems, but lots of existing products that weren't designed with HomeKit in mind will have trouble.

This is basically the big problem. Outside of caching state internally (yech), you're almost always going to have to fetch the state from somewhere. Whether that's a cloud server, a hub, or the device itself.

Making this synchronous isn't something that can possibly work well, again, outside of keeping a cache. And caching goes out the window in the case of changing state via different methods. I control my lights from a dashboard as well as Siri (HomeKit)

Let's take the Wink platform as an example. Wink works like this:

`request to server` :arrow_right: `server tells your hub to make a change` :arrow_right: `hub tells devices to do a thing`

> Ok, you cannot get a push notification on status changes?

Where would this come from? And where would it go to? I'm confused about this part.

>I'd like to look through the existing accessories supported by homebridge to see how many of them would be able to do synchronous onRead.

I suspect not many :(

snowdd1:
>  >  Ok, you cannot get a push notification on status changes?

>  Where would this come from? And where would it go to? I'm confused about this part.

I am working with an [home automation bus system](http://www.knx.org/knx-en/index.php) - every light switch on my walls is wired to that bus. whenever I press a light switch a "telegram" is sent over this bus wire, to the [electric actuator](http://www.gira.com/en/gebaeudetechnik/systeme/knx-eib_system/knx-produkte.html?t_id=6) that changes the 110/230V line status, and is eventually caught by my raspberry based bus monitor device. With a fitting event handler for connections to that, I get an event fired for each light switch status change. The same happens if I do a change via my home server's visualization, as - in the end - all effective changes have to be sent over the bus wire.
This theoretically applies to thermostat values, actuals as well as target, roller blind positions, wind speeds and rain detector, and so on.

I learn from your question that most of the other "smart home" devices do not provide active feedback on their status or metering. In that case, the home kit app cannot actively be informed about changes and will - local caching on the device and in the cloud - show an outdated status until you ask Siri for an update. As an Mac illiterate, I have downloaded the one available homekit app at that time - myTouchHome - which doesn't even have a button for refreshes (and is missing to advertise its event enablement, but that is a different story)

As I just mentioned in [nfarina's issues](https://github.com/nfarina/homebridge/issues/6#issuecomment-115372708) the problem might not be the synchronous/asynchronous topic, but the fact that while waiting for the results the http header must be sent to Siri for her to wait for the remainder of the answer.

maddox:
> As I just mentioned in nfarina's issues the problem might not be the synchronous/asynchronous topic, but the fact that while waiting for the results the http header must be sent to Siri for her to wait for the remainder of the answer.

This makes sense. I"m not sure exactly how to do this. I can take a look though. 

For what it's worth, the timing is working fine for me (as you can see in the screenshots above), and I'm using Wink, which is notoriously slow, lol. I know that's not definitive and is practically meaningless while trying to make something reliable, but at least it's anecdotal.

PR #86, opened 2015-06-26
Async onRead

maddox:
:construction: :construction:  Don't merge yet. Needs discussion :construction: :construction: 

Ok, this is try number 2 to get async state checks for accessories. These are fresh commits since the last ones were actually merged, but then reverted.

## The Goal

This adds callbacks to the `onRead` function, so that we can have asynchronous calls to fetch the state of accessories. HAP-NodeJS might not be the only pace that accessories/devices will be controlled from. So this means HAP-NodeJS won't always have the actual current state. 

Because of this, we need to somehow fetch the current state of an accessory. To do this, we'll need to use callbacks to return the values when we've fetched them. That's what this pull enables.

[According to some research](https://github.com/nfarina/homebridge/issues/6#issuecomment-115495434) by @nfarina, Siri times out at 10 seconds. This should not only be more than enough time to fetch state and return it, but it proves that HomeKit is prepared to wait for this answer. So I think this is more than fine to do this way.

### No Hack to Handle nil `onRead`
This time around, there's no check to make sure `onRead` is ready to be read. `onRead` is properly wired up in the initializer in `Core.js` and `BridgedCore.js`. So the first call to `valueForUpdate` won't fail. So we don't need to handle that weird issue (thank god, that code was embarassin : )

### Bridged Accessories

These callbacks are supported in Bridged Accessories now.

### Handle nil `onRead` Case

It is now handling the case where onRead hasn't been configured by an accessory.

### Notes

Ok, a lot has been said on this topic in the last couple of days, across a lot of threads, heh. Let's try to start over with this pull and keep the discussion in a single place. Thanks for everyone's :eyes:s on this! Please let me know anything  you see that is a problem or would make this better.

/cc @KhaosT @snowdd1 @nfarina

### DIFF
--- a/BridgedCore.js
+++ b/BridgedCore.js
@@ -56,7 +56,7 @@ for (var i = 0; i < accessoriesJSON.length; i++) {
 		//loop through characteristics
 		for (var k = 0; k < accessoriesJSON[i].services[j].characteristics.length; k++) {
 			var options = {
-				type: accessoriesJSON[i].services[j].characteristics[k].onRead,
+				onRead: accessoriesJSON[i].services[j].characteristics[k].onRead,
 				type: accessoriesJSON[i].services[j].characteristics[k].cType,
 				perms: accessoriesJSON[i].services[j].characteristics[k].perms,
 				format: accessoriesJSON[i].services[j].characteristics[k].format,
--- a/BridgedCore.js
+++ b/BridgedCore.js
@@ -56,6 +56,7 @@ for (var i = 0; i < accessoriesJSON.length; i++) {
 		//loop through characteristics
 		for (var k = 0; k < accessoriesJSON[i].services[j].characteristics.length; k++) {
 			var options = {
+				type: accessoriesJSON[i].services[j].characteristics[k].onRead,
 				type: accessoriesJSON[i].services[j].characteristics[k].cType,
 				perms: accessoriesJSON[i].services[j].characteristics[k].perms,
 				format: accessoriesJSON[i].services[j].characteristics[k].format,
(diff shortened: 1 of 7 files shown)


nfarina:
Wow, our timing is impeccable ;)

nfarina:
In fact I think our PRs are literally identical...

maddox:
lol, its amazing.

BridgedCore.js
```diff
@@ -56,6 +56,7 @@ for (var i = 0; i < accessoriesJSON.length; i++) {
 		//loop through characteristics
 		for (var k = 0; k < accessoriesJSON[i].services[j].characteristics.length; k++) {
 			var options = {
+				type: accessoriesJSON[i].services[j].characteristics[k].onRead,
```
nfarina: I think `type` is duplicated here

maddox: Good catch.

nfarina:
Alright well, see my comments from #87 for my view on this PR :) I think it's a good one!

KhaosT:
I tied and it seems working fine ^^

maddox:
Did I manage to handle all the missing issues I didn't previously? Promise it won't get reverted? Asking for a friend over on nfarina/homebridge#35.

nfarina:
I compared our respective clean-room implementations in detail and I couldn't find any significant differences besides the one typo. I also tested with Core.js, BridgedCore.js, the GarageDoor sample, etc. I think it's pretty solid!

maddox:
WOOT. Thanks everyone!

rayzorben:
Added an onRead to my device, and implemented it. I do have a physical device to test with. I have tried light bulbs and ceiling fans, and I simply never see the onRead method being called.

maddox:
@rayzorben: this pull has been merged. Could you open a new issue with your question and we can discuss it there? Thanks!

PR #87, opened 2015-06-26
Fix onRead, make async

nfarina:
This PR attempts to re-do async support as well as fixing the existing `onRead` support. See other discussions in [HomeBridge #6](https://github.com/nfarina/homebridge/issues/6#issuecomment-115417438) and #82.

I tried to take @maddox's original async branch and simplify it such that the changes to the `HAP-NodeJS` codebase are as minimal as possible, while also including his fix for `BridgedAccessoryController.js`.

I think that in a perfect world all accessories would be able to receive proactive updates about their state, either by pushed local network events from the accessory itself or from webhooks in the cloud. But many existing accessories were simply not designed for HomeKit and implementing realtime updates for the purposes of synchronous state reading is impractical.

HomeKit is asynchronous by nature and Siri has been designed with up to 10 seconds of "patience" for learning about accessory state. So I think it is reasonable to allow accessory implementors to do a little bit of "work" before responding to her. This doesn't work for 100s of lights but it works great for single accessories like Lockitron.

The changes to `HAP-NodeJS` to support async reads are pretty minimal, just passing a couple callbacks along from Server to Characteristic. I think it's pretty straightforward.

In practice, this is working really well for me - I've implemented `onRead` for Lockitron and it's very quick to check state from their API. Siri now properly says "Your door locks are locked."

/cc @maddox, @snowdd1

### DIFF
--- a/AccessoryController.js
+++ b/AccessoryController.js
@@ -43,19 +43,20 @@ AccessoryController.prototype = {
 		}
 		return JSON.stringify(dict);
 	},
-	jsonForCharacteristicUpdate: function jsonForCharacteristicUpdate(aid, iid) {
+	jsonForCharacteristicUpdate: function jsonForCharacteristicUpdate(aid, iid, callback) {
 		var charObject = this.objects[iid];
-		var charValue = charObject.valueForUpdate();
-		var respDict = {
-			characteristics: [
-				{
-					aid: aid,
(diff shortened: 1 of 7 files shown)


nfarina:
OK closing this as @maddox is proposing the exact same changes in #86! Let's discuss there instead.

PR #102, opened 2015-07-20
Added humidity, temperature stypes

jmtatsch:
for use with homebridge and the upcoming FHEM platform shim http://forum.fhem.de/index.php/topic,32652.msg314166.html#msg314166
Also more sensor = more fun ;)


justme-1968:
window would be great.

for the other types: it would be great to know where they are from. i can not find them in the apple headers. if is use output volume in eve it looks like a custom field.

Issue #109, opened 2015-08-02
rename window characteristics


justme-1968:
the three characteristics WINDOW_COVERING_TARGET_POSITION_CTYPE, WINDOW_COVERING_CURRENT_POSITION_CTYPE and WINDOW_COVERING_OPERATION_STATE_CTYPE should probable renamed to just WINDOW_... (without the COVERING_) as these are also used by the window service class to open and close windows.

i'dont know if they are already used and it would be a problem for backwards compatibility.

the window service type would be: exports.WINDOW_STYPE = stPre + "8B" + stPost;

nfarina:
This should be closed by #110 (all known services/characteristics are extracted from HomeKit Simulator and provided in HomeKitTypes.js).

Issue #85, opened 2015-06-26
JS class style?


nfarina:
HAP-NodeJS uses a style of `require`ing and `new`ing objects that I haven't seen before. Each module exports a dictionary of constructors, and modules that use constructors to make classes use a "factory" (?) pattern:

```js
var characteristic_Factor = new require("./Characteristic.js"); // new is used here
[later]
var characteristic = new characteristic_Factor.Characteristic(...); // and here
```

Meanwhile, classes seem to "hijack" their initializer in a strange way:

```js
function Characteristic(options, onUpdate) {
	if (!(this instanceof Characteristic))  {
		return new Characteristic(options, onUpdate); // repeated init options
	}
	... rest of normal init ...
```

I'm curious if there is an advantage to this pattern versus "normal" javascript classes?

KhaosT:
@nfarina do you have an example of "normal" javascript class? I'm pretty new to JavaScript and that's the way I found people use in other repos...

nfarina:
Gotcha! Totally understand. I would personally copy from a really popular NodeJS lib like Express, it uses the traditional JS class structure and exposes the constructor function directly: https://github.com/strongloop/express/blob/master/lib/router/route.js

nfarina:
Would you mind if I submitted a PR to for this? I also might take a stab at documenting some of the classes in the headers (and you can tell me when I guessed wrong).

maddox:
Maybe let #86 land first tho? :sweat_smile:

maddox:
Docs would be :heart::heart::heart:

KhaosT:
@nfarina feel free to do so :)

snowdd1:
> Meanwhile, classes seem to "hijack" their initializer in a strange way:
> 
````
function Characteristic(options, onUpdate) {
     if (!(this instanceof Characteristic))  {
         return new Characteristic(options, onUpdate); // repeated init options
     }
     ... rest of normal init ...
`````
> 
> I'm curious if there is an advantage to this pattern versus "normal" javascript classes?

[...]

PR #110, opened 2015-08-03
HAP-NodeJS New API, Protocol Refactor

### LINKED ISSUE #88: Exception when wrong Pin is entered.
Is there a way to have it gracefully handle a bad pin? Right now it just throws an exception and stops execution. Would be nice to have it remove the device from persist and recreate it in this case to make it available for discovery again.

nfarina:
Closes #88, #109, #85

This is a pretty massive PR, I'm aware.

Inside is a significant reworking of HAP-NodeJS internals, as well as the introduction of a new API for library consumers.

### Reworking

The reworking was done to achieve a few goals. The first and most important is a stable Bridge accessory that can handle adding/removing Accessories over time without corrupting HomeKit's database on iOS. This enables a superior experience for the end-user in that they only need to pair with the Bridge and enter one PIN code, then the server can later add and remove Accessories all day without the user needing to do anything.

The other goal was to create an organized, encapsulated, and well-documented codebase that encourages more contribution from other developers in the community. Classes do not share state directly, and communicate using an event-based model. And each major class has a large header attempting to explain why and how it does what it does.

### Library API

The final goal was to create a complete and easy-to-use API for developers writing Accessories. The existing "object literal" notation still works, but you can also now create Accessories, Services, and Characteristics in code using an extensive set of subclasses automatically extracted from HomeKit Accessory Simulator.

See the [README](/nfarina/hap-NodeJS/tree/hap-server-refactor) file in this PR for more details.

There shouldn't be any breaking changes, but note that because of changes to the storage model in `persist/`, your existing pairing relationships will not carry forward. You will need to re-pair with your devices when running this version for the first time.

### Changelog

  * Restructure server implementation from the ground up
  * Server remembers HomeKit IDs for stable Bridged mode
  * New EventEmitter-based library API
  * Characteristic get/set can be async
  * Automatic importing of all known Services/Characteristics
  * Correctly handles invalid pincode
  * Event subscriptions tracked per-connection
  * Events muted for originating channel
  * HAP "configuration flag" correctly incremented on config change
  * Sends periodic "null events" to keep long-lived HAP connections open


volschin:
Sounds very interesting.
Do you have a compatible homebridge branch ready to do some more testing? 
The actual master is not compatible, because of function changes.

nfarina:
Working on that next!

nfarina:
OK there's now a branch of HomeBridge that uses this WIP version of HAP-NodeJS: https://github.com/nfarina/homebridge/tree/use-hap-refactor

Just remove the old HAP-NodeJS from `node_modules` and do an `npm install` to fetch this version.

snowdd1:
Wow, that's impressive. 
I have to admit I did not yet grok all of it!
I found a method missing I was using so far: characteristic.updateValue(value, peer) to update the cached value when an external event changes a status of a characteristic, such as a hardware light switch being pressed. How is that supposed to be handled with the new characteristic class? Method setValue(value, callback, context) is the right thing? What is context supposed to be?
Thanks
Snowdd1

nfarina:
Hm that is a good question - calling `setValue` would do the trick, as it updates the cached value and emits the `change` event which notifies any iOS listeners. However it also emits the `set` event as if it were iOS attempting to set the value instead of you yourself. So it's a bit confusing. Perhaps `setValue` should only be for updating state yourself, and we need new methods for iOS-initiated actions?

nfarina:
Oh and `context` is not something that you'd normally need to use - it's a mechanism to preserve the caller of the `setValue` function such that the eventual corresponding `change` event doesn't propagate back to the iOS device that initiated it. (see notes in Accessory.js:411 if you're curious)

nfarina:
Followup idea for the `setValue` question - if `context` were actually passed back to the `set` function as well as `change` then you could leverage `context` yourself to make sure that you didn't act on your own `setValue` events.

snowdd1:
Yes, the only thing we do with such changes would be informing other peers (iOS devices) of the change that happened, if their HomeKit apps are event enabled.
Still I have to do much reading into your code...

snowdd1:
> Perhaps setValue should only be for updating state yourself, and we need new methods for iOS-initiated actions?

Or... what about having a setCachedValue() method for handling device status changes that are actively advertising their updates? For the bus systems I wrote about...

KhaosT:
So far it's working pretty good, thanks for all the efforts :) One small issue, since iOS will establish a single connection with accessory (to handle event triggers), we should disable the timeout of our internal http server. Right now I suggest to set ``this._httpServer.timeout = 0;`` in eventedhttp.js after creating the internal http server ^^

nfarina:
Oh that is great, I didn't realize Node's HTTP server had a default timeout! Will fix that now.

nfarina:
Quick followup from testing - I've had my server running this code on a Raspberry Pi for a couple days now with no issues - serving up about a dozen accessories behind a single Bridge. It's pretty great to add and remove accessories on the server and instantly be able to use them with Siri :)

nfarina:
Just got the same mysterious crash as described here: https://github.com/nfarina/homebridge/pull/94#issuecomment-129111092

Need to investigate further.

justme-1968:
i have a problem that is probably related to the timeout = 0.

after a few hours or days depending how often i leave the house or switch the network on the iOS device on and off i get lots of stale open connections to the web server ports. with the old version 200-300 and then homebridge stops accepting commands from the iOS device.

justme-1968:
i have changed the timeout to 1 hour and it looks much better now. just a few open sockets and if i try to with in eve then any closed ones are just reopened.

cflurin:
Testing with the Debugger (DEBUG=* node app.js) you will see an endless loop:
>Accessory [HomeBridge] Getting Characteristic "Manufacturer" value  +0ms

The reason ist because the var "value" is null.
To avoid it I changed the code in Accessory.js:

    Accessory.prototype._handleGetCharacteristic = function(aid, iid, events, callback)
    ...
    -- debug('[%s] Getting Characteristic "%s" value ', this.displayName, characteristic.displayName);
    ...
    ++ if(!value)
    ++   value = "empty";
    ++ debug('[%s] Getting Characteristic "%s" value %s', this.displayName, characteristic.displayName, value);
    
    // all done - compose the response and send it to the callback
    callback(null, {
      characteristics: [{

Just now I don't see, why the value is null.

nfarina:
How are you getting iOS to query for "Manufacturer?"

cflurin:
I use the EVE App. I've just defined one Accessory for testing purpose.
The file can be downloaded from Gist:
https://gist.github.com/cflurin/65748e04ea16fd777e6e

nfarina:
OK so here's the behavior I've observed:

  1) Run HAP-NodeJS `BridgedCore.js`
  2) Pair iOS client with the bridge
  3) iOS client connects to HAP-NodeJS, leaves connection open
  4) Now I manually shut off Wifi on iOS device
  5) Server does nothing; socket object receives no events - no 'end', no 'timeout', no 'close', etc.
  6) Wait ....
  7) Turn iOS device Wifi back on
  8) If not very much time has passed, the connection resumed; one connection remains open
  9) If more time has passed (minutes), iOS opens up a new connection - there are now TWO connections open on the server, as the first one has still not been closed

I'm wondering if, even though we set timeout to "infinite", if there is an underlying unix TCP connection timeout that is causing this unclean `ECONNRESET` shutdown - from some light Googling, it seems there may be a hard system limit of 24-72 hours for an underlying socket timeout. That would mesh with my experience so far - takes about 2 days on my RPi for this `onread` exception to happen.

cflurin:
As I described above I've only one Accessory.
But threre are two aids (1 and 2), I suppose the first is for the HomeBridge and the second for my device (flex_lamp).
In the iOS App settings I can see both Accessories (HomeBridge and flex_lamp)
The HomeBridge Manufacturer and the Model are undefined otherwise the Manufacturer and Model for the flex_lamp are set as defined in FhemAccessory.fs

Take a look at the pictures.

img_1154
img_1155
img_1156

justme-1968:
yes. there is a low level tcp/ip timeout of about 2 days. 

but i think we are missing something. there must be a way to recognize these stale connections. maybe a test with the homekit simulator would be good to check if it shows the same behavior of open connections.

nfarina:
Hi @cflurin, my previous response was referring to another issue with socket timeouts - but I also downloaded the EVE app and can reproduce the infinite loop you're seeing. I'm looking at that now. Would you mind resizing or removing the screenshots from your comment so this thread is a bit more readable? :)

nfarina:
@cflurin Give the latest commit a shot - it should fix the infinite loop problem. Turns out iOS _really_ wants non-null values for `AccessoryInformation` :)

cflurin:
Ok, thanks.
One small change: I'd move the debug line inside the function "characteristic.getValue(function(err, value)" so you can log the value and also modify the debug Error ...
```javascript
    //set the value and wait for success
    characteristic.getValue(function(err, value) {
  
+   debug('[%s] Getting Characteristic "%s" value %s', this.displayName, characteristic.displayName, value);
  
    if (err) {
-     debug('[%s] Error getting Characteristic "%s" value: ', this.displayName, characteristic.displayName, err.message);
+     debug('[%s] Error getting Characteristic "%s" error message %s', this.displayName, characteristic.displayName, err.message);
      callback(err);
    }
```

I'm not sure whether the comment is correct, I think it should be "get the value ..."

nfarina:
Good call - updated.

nfarina:
@KhaosT Any thoughts about the "stale" connections that can be left open for (potentially) days? I can confirm that setting `httpServer.timeout = 0` is what's causing this. (It also happens in HAP-NodeJS master). iOS doesn't seem to mind if the connection times out while it's "away" from wifi - it simply reconnects automatically.

cflurin:
@nfarina You are missing "%s" in your debug log:

```javascript
  characteristic.getValue(function(err, value) {
    
    debug('[%s] Got Characteristic "%s" value: %s', this.displayName, characteristic.displayName, value);

    if (err) {
      debug('[%s] Error getting value for Characteristic "%s":  %s', this.displayName, characteristic.displayName, err.message);
      callback(err);
    }
```

nfarina:
Ah, thanks, fixed. Technically the underlying console.log call will append the last argument but this is clearly better :)

cflurin:
One of my main issue is connecting my iPhone (just updated to iOS 8.4.1) to the HAP-Server.
If I turn off the Wifi and turn on again, I lose the connection to the server.
The same happens if I stop and restart the server.
Just now I've to soft reset the iPhone.

Any ideas?

nfarina:
Hm, this has not happened to me on iOS 9 so far (or on iOS 8.4 from what I can recall) ... do you mean this new refactored branch or master branch? And what does the server print when you run via `DEBUG=* node BridgedCore.js` after you turn off Wifi? Does it ever print anything after that, or after you turn your wifi back on?

justme-1968:
i have the timeout set to 1 hour for two days now and had no problems so far. there are no stale connections (or at least only one ours worth) and homekit/eve just reconnects automatically. (at least for all bidirectional devices. there are three thermometers that have problems but this may be completely unrelated.)

cflurin:
The problem exists since I started testing Homebridge (both master and refactor).
I don't get any logs at all (neither running via DEBUG). The HAP-Server is running on a Mac mini (OS X 10.10.4).

nfarina:
Could you verify that the issue still occurs when running the master branch of HAP-NodeJS with a clean checkout using the sample accessories? If it still happens then, I would recommend filing a separate issue in this repo and we can move the discussion there. Maybe others have had this same problem.

cflurin:
OK, I've tried again with the master branch of HAP-NodeJS and DEBUG=* node BridgedCore.js without success. I already posted this issue https://github.com/KhaosT/HAP-NodeJS/issues/108

cflurin:
... I tried connecting my iPhone with the Mac mini using the the Mac mini Wifi instead of the router wifi and it seems to work now :)

nzhome:
yep I tested this is working successfully.  although sometimes if I changed config.json then it would force me to create a new username in app.js for it to be seen again.   and it seems that you only get one shot at adding it.

cflurin:
@nfarina
I'm trying to implement an update to the iOS device like this example in the master branch:

````Javascript
{
    cType: types.POWER_STATE_CTYPE,
    // ...
    onRegister: function(characteristic) {
          characteristic.eventEnabled = true;
          that.connection.on(that.switchAddress, function(msg) {
              var newValue = msg.getData()[0] === 1;
              characteristic.updateValue(newValue, null);
          });
    },
    // ...
}
```
but I'm missing the "updateValue" function in the refactor branch.
What shall I use instead?

snowdd1:
SetValue()
See above my first comment on the PR

cflurin:
@snowwdd1 Thanks. I did read it a couple of days ago, sorry my fault.

I suppose I can used it this way:

    characteristic.setValue(value);

cflurin:
Hi,
The EVE App set the value to 0/1 in place of false/true, maybe a bug in EVE (the format should be boolean). This has an implication using setValue(value). I solved it this way:

```` javascript
onUpdate: function(value) {
          value = value ? true: false;
          if(value != this.currentState)
            this.setPowerState(value); 
        }.bind(this),
````

Insteon+ set the value to false/true.

nzhome:
it is not a bug in EVE, the program will have to handle all 4

nfarina:
@justme-1968 @cflurin Just added a commit that attempts to detect closed connections (i.e. turn off Wifi on iOS) instead of leaving them open for a long time. If you have the ability to test this that would be really helpful!

cflurin:
@nfarina  I've just updated to the new HAPserver.js

1. Turning off Wifi on the iOS and on again works, now we have to wait at least one day.
2. Restarting homebridge ist OK too.

I'm working on news Accessory shims and I've to intensely restart homebridge, all this without any issue :-)

nfarina:
> Turning off Wifi on the iOS and on again works, now we have to wait at least one day.

Do you mean that if you toggle Wifi on your iOS device, it won't reconnect until the next day? In my experience it usually reconnects immediately…

KhaosT:
After using the branch for the past few weeks, I believe it's stable enough to merge into the master. @nfarina Thanks for all the efforts :)

cflurin:
@nfarian: To precise, I mean there are two different cases, reconnecting my iPhone works immediately. The second case is after more or less one day. It seems the connection get lost after this time period.

nfarina:
:fireworks:

PR #120, opened 2015-08-24
Support Service display names

nfarina:
- Convert sample Light accessory to new API


KhaosT:
Looks Nice. One small problem, HomeKit suggests developer to use "name" characteristic to decide if the service should be displayed to user... We shouldn't include "name" characteristic in all services.

nfarina:
Ok let me think on this a bit...

nfarina:
(force pushed) is that more as it should be?

nfarina:
Added some examples of service names in Light/Lock.

PR #123, opened 2015-08-29
Protect certain callbacks against multiple calls

nfarina:
Especially for Characteristic get/set events, we want to make sure that
calling them twice results in an Error (as this will otherwise cause
unpredictable problems).


KhaosT:
Just curious, under what kind of circumstance will we call setter/getter twice?

nfarina:
We shouldn't—it would be considered "programmer error"—but developers making accessories may accidentally call them twice for various "async is hard" reasons :) (I've seen it already)

Issue #121, opened 2015-08-24
How do I access optionalCharacteristics?


volschin:
In Services.js is specified
`this.optionalCharacteristics = [];`
and filled from HomeKitTypes.js. A requied characteristic can be accessed by `getCharacteristic` or `setCharacteristic`. How this has to be done for the optional ones?

A small example:

``` javascript
var sensorCO2Service = new Service.CarbonDioxideSensor();

sensorCO2Service
  .getCharacteristic(Characteristic.CarbonDioxideDetected)
  .on('get', 'true');
  .getCharacteristic(Characteristic.CarbonDioxideLevel)
  .on('get', function(callback) {
    // return our current value
    callback(null, FAKE_SENSOR.getCO2());
  });
```

`CarbonDioxideDetected` works, `CarbonDioxideLevel` doesn't.

KhaosT:
Try
```js
sensorCO2Service
  .getCharacteristic(Characteristic.CarbonDioxideDetected)
  .on('get', 'true');
sensorCO2Service
  .getCharacteristic(Characteristic.CarbonDioxideLevel)
  .on('get', function(callback) {
    // return our current value
    callback(null, FAKE_SENSOR.getCO2());
  });
```
?

cflurin:
Http.js shows how to add an optional Characteristic:

````javascript
    lightbulbService
      .addCharacteristic(new Characteristic.Brightness())
      .on('set', this.setBrightness.bind(this));
````

KhaosT:
@cflurin yeah, you're right... just found that from samples :)

volschin:
Works, thank you. :+1:

volschin:
I reopen this. Using addCharacteristic is the same way it is used in HomeKitTypes.js for required characteristics. You have no added value, no consistency check, that your added characteristic is allowed in this context.

I think there should be a user function for those optional characteristics, that check this is an allowed one with the data from `this.optionalCharacteristics = [];`.

snowdd1:
@volschin I think you're right. 
@nfarina What happens to the `optionalCharacteristics` that are pushed into that array of the service definitions in *HomeKitTypes.js* ? Are the used somewhere? If not, volschin's idea of using them as a security check (issue a warning to the log/console e.g.) might prove helpful.

nfarina:
[...]

PR #128, opened 2015-09-11
Get characteristic security enhancements

snowdd1:
As discussed in #121 

Service.getCharacteristic(name, nofail) fails with an exception if name is not an existing or optional Characteristic. If it is an optional, it gets added and returned.
For internal use, nofail will return undefined if not found. Allows for getCharacteristic(Some, true) || addCharacteristic(Some, true).

Service.addCharacteristic(Characteristic.Type, nowarn) issues a warning to console if the characteristic.Type is not found in optionalCharacteristics of the service. nowarn=true suppresses this warning (if you know what you are doing, and of course for initial creation of service types)

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -152,7 +152,7 @@ Accessory.prototype.getService = function(name) {
       return service;
   }
   // failed
-  throw new Error("Accessory.prototype.getService: failed to find service "+ (typeof name === 'string') ? name : name.UUID )
+  throw new Error("Accessory.prototype.getService: failed to find service "+ (typeof name === 'string') ? name : name.UUID );
 }
 
 Accessory.prototype.addBridgedAccessory = function(accessory) {
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -151,6 +151,8 @@ Accessory.prototype.getService = function(name) {
     else if (typeof name === 'function' && service instanceof name)
       return service;
   }
+  // failed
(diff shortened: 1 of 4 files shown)


nfarina:
So I think this is a great start, but a couple things I'd be cautious about. I'm not super into the extra Boolean argument, it seems a bit fiddly for functions that are already a bit too "magic". Also I was thinking this change would be more about better error messages and throwing errors when attempting to access a characteristic that you haven't added yet. I think it's better to force the user to add the characteristics they want and not auto-create-them-sometimes in the getter.

snowdd1:
> I'm not super into the extra Boolean argument

Me neither, but I encountered constructions like `getCharacteristic(xxx) || addCharacteristic(xxx)` - if you wanted to keep these you need to tell it not to fail but to return undefined
And the error message came up when initially creating the templates, as the optionals were not there at that time. Probably just reverse the order, add all possible types to the optionals before the `addCharacteristic()` of the required c.s? 


> I think it's better to force the user to add the characteristics they want and not auto-create-them-sometimes in the getter

Actually that's the part @KhaosT liked the most, and me too. It makes it really ease to add characteristics, as long as they're in the optionals array.

Yet I am not convinced I fond the best way to generate the warnings when creating a characteristic that is not optional to the service - quite some issues here and in HAP-nodeJS deal with *why Siri doesn't react to my XYZ characteristic*?

snowdd1:
As this discussion is abandoned I am closing this for now.

snowdd1:
Replaced by #144

PR #152, opened 2015-10-22
Updated Service.js

henzed:
You are looping through the "optionalCharacteristics", but trying to read at the index of the normal "characteristics". Therefore it will always return "undefined" when you request optional characteristics via "getCharateristics".


KhaosT:
Thanks

PR #176, opened 2015-12-08
Save context created by plugins

cflurin:
As far as I know context was originally introduced in order to be able
to save plugins (shims) properties like who did trigger "setValue()".
This modification prevents that context is overwritten by hap-nodejs.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -437,7 +437,7 @@ Accessory.prototype._handleGetCharacteristics = function(data, events, callback)
     // cached Characteristic value, an internal 'change' event will be emitted which will cause us to
     // notify all connected clients about that new value. But this client is about to get the new value
     // anyway, so we don't want to notify it twice.
-    var context = events;
+    var context = null;
 
     // set the value and wait for success
     characteristic.getValue(function(err, value) {
@@ -511,7 +511,7 @@ Accessory.prototype._handleSetCharacteristics = function(data, events, callback)
     // by Characteristic and passed on to the corresponding 'change' events bubbled up from Characteristic
     // through Service and Accessory. We'll assign it to the events object since it essentially represents
     // the connection requesting the change.
-    var context = events;
+    var context = null;
     
     // if "ev" is present, that means we need to register or unregister this client for change events for
     // this characteristic.
--- a/lib/Characteristic.js
+++ b/lib/Characteristic.js
@@ -46,6 +46,7 @@ function Characteristic(displayName, UUID, props) {
   this.UUID = UUID;
   this.iid = null; // assigned by our containing Service
   this.value = null;
+  this.context;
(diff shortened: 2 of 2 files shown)


KhaosT:
The context in hap-nodejs is used to prevent sending repeated notifications to iOS device which started the changes, it's not for plugin to save who initiate the event. By setting context to null, the hap server will send unnecessary notifications to iOS device.

This PR will not be merged as HomeKit Accessory Simulator explicitly prevent server to send notification to iOS device which started the change.

cflurin:
@KhaosT: obviously I didn't consider that context is also used by the hap-server.

The purpose of my proposal is to assure that the context can be saved. I'd like to use context to save properties like timestamp and trigger i.e. who did use setValue(). As I noticed this is also already used by  plugins developers.

I propose another solution (in Characteristc.js) that shouldn't impact the hap-server.

```
  if (context && !context.keepalive) {
    this.context = context;
  }
  else {
    context = this.context;
  }
```
The assignment in Accessory.js will be unchanged

```
    var context = events;
```
What do you think about?

KhaosT:
@cflurin that looks better. Just curious, why would be prefer to save context on the hap-nodejs's characteristic object itself instead of having developer to save it somewhere else in their model? Also, since developer can already add extra objects onto the characteristic object itself if they really needed to, why do you think it's necessary to standardize the "extra" context?

cflurin:
Probably there are other solutions but this seems to be quite simple. 

context ist used in the plugin to prevent infinite loops. As you know, setting a value using setValue() in the plugin will also emit the event 'set'. This way the plugin can check context and avoid an infinity loop. I'm thinking about to save properties in context that I can use when I get events like 'set' or 'characteristic-change'.

For more details please take a look at the homebridge-punt repository.

KhaosT:
setValue takes in a context, plugin should use that in cases like this and doing so should be sufficient enough to prevent infinity loop.

cflurin:
That's right, for the "infinity loop" case. I'd like to save properties like timestamp in the context. Just now this is lost as soon the 'set' or 'get' event is emitted. Sure I can try to find another solution if the modification in hap-nodejs isn't suitable.

cflurin:
Right save_andsetValue() just now store two properties (timestamp and trigger). I use the properties for the WUI puntView. Unfortunately as soon the hap-server uses context I get a context like {"keepalive": true ...

If you like, you can install homebridge-punt and use the build-in simulator without connecting to the fhem-server.

KhaosT:
@cflurin hap-nodejs will only send your plugin context like that if a request is initiated by iOS device (or plugin invoked the setValue with the same context passed from event). Based on the description and what I saw from homebridge-punt, this is a case that can be handled by the plugin by improving data model to handle the value changes.

PR #204, opened 2016-01-29
Prevent clearing AccessoryInfo files

mschm:
persistSync() should not be called, before seItem() is ready. In my case, AccessoryInfo files were cleared, which caused recreation and ultimately preventing reconnection. Calling synchronous setItemSync() fixes that.


KhaosT:
@mschm thanks for fixing this :) Would you mind to also replace the setItem at [here](https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/model/IdentifierCache.js#L122) and I'll merge them in one PR.

KhaosT:
Awesome :)

mschm:
Thanks!!

PR #205, opened 2016-01-31
added maxLen validation for string characteristics

thallth:
otherwise iOS complains and about half of the accessories won't appear


KhaosT:
Thanks :)

PR #212, opened 2016-02-04
Change deprecated 'licenses' property to 'license'

htreu:
'npm install' warns about no license field available in package.json.


KhaosT:
That's a strange NPM requirement ...

Issue #131, opened 2015-09-16
Ports 


volschin:
In eventedhttp.js a port is hardcoded, that prevents to run more than one instance of homebridge on a machine.

``` javascript
var PortAllocator = {
  initialPort: 54826, // first port to use; subsequent ports will increment from this
  portsInUse: {}, // sparse array of ports in use
```

Is there a technical reason to have this as fixed port or could it be made configurable?

My intention is to run a more productive and a test homebridge on one machine.
I found that problems with a bridge device need to remove all other devices or reset of complete homekit necessary. Having more than 50 devices this makes me not lucky.

nriley:
Even if there is only one copy of HAP-NodeJS running it is possible for one of the ports to already be in use (in my case it is VMware NAT).  eventedhttp.js should check if the port is in use by another process prior to trying to listen.

snowdd1:
I was thinking of having the eventedhttpserver check for the availbility of te port, as in the node server documentation at https://nodejs.org/docs/latest/api/net.html#net_server_listen_port_hostname_backlog_callback

snowdd1:
@nriley : I think your approach doesn't do much better: "A port value of zero will assign a random port." is said in the documentation. I think that random is only slightly better than incremental without checking for the error element. I'll try a pull request for @KhaosT  based on the documentation example linked in mylast post

nriley:
Good point, I did not realize the OS does not check if the port was in use.  What a mess.  http://stackoverflow.com/questions/9901043/

snowdd1:
I wrecked my local git clone so I cannot produce a pull request right now. Yo might want to try the following:

https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/util/eventedhttp.js#L125 :
````JavaScript
  this._httpServer.listen(this._httpPort);
  this._httpServer.on('error'), this._onHttpServerError.bind(this);
````

https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/util/eventedhttp.js#L193 insert:
````JavaScript
EventedHTTPServerConnection.prototype._onHttpServerError = function(e){
	  // we got an error during opening of the port?
	  if (e.code === 'EADDRINUSE') {
		  debug("Port [%s] in use, trying next port", this._httpPort);
		  this._httpPort = PortAllocator.allocatePort();
		  //retry: 
		  this._httpServer.listen(this._httpPort);
	  }
}
````
[...]

PR #218, opened 2016-02-20
EventedHTTPServer: Don’t assume high-numbered ports aren’t in use (#131).

nriley:
This version actually tries a different port if the OS-generated port is in use...

### DIFF
--- a/lib/util/eventedhttp.js
+++ b/lib/util/eventedhttp.js
@@ -104,7 +104,6 @@ EventedHTTPServer.prototype._handleConnectionClose = function(connection) {
  
 function EventedHTTPServerConnection(clientSocket) {
   this._remoteAddress = clientSocket.remoteAddress; // cache because it becomes undefined in 'onClientSocketClose'
-  this._httpPort = PortAllocator.allocatePort(); // grab the next open port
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks ^^ Would you mind to bring this change also to the [external HAP Server](https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/Accessory.js#L468) in the same PR so that user will no longer need to worry about ports at all?

snowdd1:
@KhaosT : Could you explain the use of the "internal" http Server in short? I thought it would be used for communication with the (ios) Clients as well!
Thanks

KhaosT:
@snowdd1 HomeKit IP Accessory is basically just a normal HTTP Server that handles request. The only thing special is that once we finished ``pair-verify``, the protocol enforces a secure session that requires encrypt the underlying tcp socket communication. I wasn't able to find a way to modify the underlying socket before the data reaches node's http server so we ended up with a two layer structure that internally we use node's http module to host the HAP Server (plain text) and externally we use a normal tcp server (with net module) to relay request and encrypt/decrypt data to HAP Server (internal)

KhaosT:
alright, I'll do the external server part :) Thanks for the PR ^^

nriley:
Sorry I've been busy! Appreciate it.

PR #239, opened 2016-04-17
Allow to getService by subtype string

SphtKr:
Okay if there’s a reason not to do this, just seemed like it would be handy. I’m going to iterate `services` in client code to do the same thing. I suppose this could potentially be a breaking change, even with it as the last `||`, but probably very unlikely that it would break anything.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -189,7 +189,7 @@ Accessory.prototype.getService = function(name) {
   for (var index in this.services) {
     var service = this.services[index];
     
-    if (typeof name === 'string' && (service.displayName === name || service.name === name))
+    if (typeof name === 'string' && (service.displayName === name || service.name === name || service.subtype === name))
       return service;
     else if (typeof name === 'function' && ((service instanceof name) || (name.UUID === service.UUID)))
       return service;


KhaosT:
@SphtKr can you elaborate on why you are trying to get the service back using subtype instead of display name?

SphtKr:
Maybe that's what I should be doing, but here was my thought process:

1. There's no (technical) requirement for `displayName` to be unique, but `subtype` is required to be unique by design.
2. `displayName` is intended to be seen by humans and might be localized/internationalized somehow, while `subtype` should never be seen by humans and therefore does not need to be localized.
3. Anecdotally, in iOS apps, I'm not getting the result I'd expect from setting the `displayName` of a `Service`. I haven't played with it enough to figure out why yet, but for example Eve seems to make an odd choice about which name to display first (see attachment).

That last one could be something I'm doing wrong, and this isn't that important, a pull request was just the easiest way to ask the question. Thanks!

img_1529

KhaosT:
Since this one is unlikely to break anything and is indeed useful in some cases, I'll merge it ^^

SphtKr:
Cool, thanks again!

KhaosT:
alright, it's in hap-nodejs 0.3.1 now

PR #279, opened 2016-08-30
replaces node-persist by persist-json

hobbyquaker:
Replace node-persist by persist-json. This provides a more convenient way to store data in an OS-specific reasonable path by default and prevents issues when the user running hap-nodejs has no write-access to the node_modules/node-persist folder. 
Will by default create a Folder ~/.hap-nodejs/ on Linux and ~/Library/Preferences/hap-nodejs on macOS.

### DIFF
--- a/BridgedCore.js
+++ b/BridgedCore.js
@@ -1,15 +1,11 @@
 var fs = require('fs');
 var path = require('path');
-var storage = require('node-persist');
 var uuid = require('./').uuid;
 var Bridge = require('./').Bridge;
 var Accessory = require('./').Accessory;
 var accessoryLoader = require('./lib/AccessoryLoader');
 
 console.log("HAP-NodeJS starting...");
-
-// Initialize our storage system
-storage.initSync();
 
 // Start by creating our Bridge which will host all loaded Accessories
 var bridge = new Bridge('Node Bridge', uuid.generate("Node Bridge"));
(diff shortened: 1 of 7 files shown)


KhaosT:
Thanks for this PR.

Changing the underlying storage library is a little bit trick without a way to migrate existing data to new location. Since a lot people already spent a while setting up their homes, accessories and triggers, I'm afraid we cannot simply swap the persist library.

hobbyquaker:
oops, sure, you're right. didnt think about this...

hobbyquaker:
what do you think - should i invest some
time into a migration routine or do you prefer keeping the node-persist storage?

hobbyquaker:
just thought about another approach: keeping node-persist but giving the user the ability to overwrite the default save/load function with his own, that would also suite my needs and should have no negative impact for other users. so everybody could use the storage that he wants, that would help me in my project https://github.com/hobbyquaker/homekit2mqtt and also another project i started working on (both depending on your hap-nodejs)

KhaosT:
May I ask why you want to change this...? If you just want to change folder hap-nodejs save the data, you can simply invoke ``init()`` with path and hap-nodejs will save things to that path...

hobbyquaker:
The idea to completely change the storage method is mainly motivated by the ioBroker project, there is a homekit plugin that depends on your hap-nodejs also. ioBroker offers his plugins it's own pluggable storage engine and can be used with e.g. a CouchDB backend. Would be nice to be able keep all data including hap-nodejs's in one place there.

KhaosT:
I'm fine with this approach as long as it still supports old data :) feel free to send PR and I'll merge that.

PR #292, opened 2016-09-25
Add IP Camera Support

KhaosT:
Add support for IP Camera


lib/Camera.js
```diff
@@ -0,0 +1,213 @@
...
+};
+
+function Camera() {
+  this.services = [];
+  this.streamControllers = [];
+
+  this.pendingSessions = {};
+  this.ongoingSessions = {};
+
+  let options = {
```
maddox: Is there a new minimum version of Node this needs now?

Simply pulling `0.4.0` into Homebridge and launching it, this line crashes for me.

```
  let options = {
  ^^^
SyntaxError: Unexpected strict mode reserved word
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/Users/jmaddox/source/OSS/homebridge/node_modules/hap-nodejs/index.js:3:14)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
```

PatchworkBoy:
Node 4.3.2, or, earlier versions with the Harmony flag set. See https://nodejs.org/en/docs/es6/

lib/Camera.js
PatchworkBoy: Node 4.3.2, or, earlier versions with the Harmony flag set. See https://nodejs.org/en/docs/es6/

maddox: 👌

KhaosT: @maddox @patchworkboy do you think it's necessary to support at least node 4.0.0? Or is it fine to support just newer version?

PatchworkBoy: Put it this way... I'M struggling to upgrade from v0.12.x and get everything up and running again. Just dropped v6.6 on and after updating homebridge to latest, looking at:

> /usr/local/lib/node_modules/homebridge/node_modules/hap-nodejs/node_modules/mdns/lib/dns_sd.js:35
>     throw ex;
>     ^
> 
> Error: Module version mismatch. Expected 48, got 14.

I think if you’re going to force people up to a newer version of Node than 0.12.x then you’re gonna have to document the process for them... (thinking of the average homebridge + plugin user here as it’s these who are hammering down the door wanting IP Camera support, rather than folks developing for themselves with HAP-NodeJS).

It sucks, I know, but I fear it may be necessary...

KhaosT: @PatchworkBoy Module version mismatch is caused by pre-compiled modules. Try [npm rebuild](https://docs.npmjs.com/cli/rebuild)

lisergio:
Is possible conect a webcam to the raspberry pi and use cameracore.js???
Thanks!

KhaosT:
@lisergio yes, please check out ffmpeg on how to use webcam as input.

lisergio:
I'm sorry but I can not find it... ☹️

KhaosT:
@lisergio you can start from [here](https://trac.ffmpeg.org/wiki/Capture/Webcam).

PR #294, opened 2016-09-26
Update Function, to only updating the remote value but not firing the…

TheNoim:
… 'set' event.


PatchworkBoy:
I thought that was the point of having contexts in set...? ie: Just issue the set with a random context.

KhaosT:
@PatchworkBoy I guess it makes life easier by offering an extra method... otherwise if plugin developer isn't care, it can result set-loop...

Issue #322, opened 2016-11-03
Node 4 support


mifi:
https://github.com/KhaosT/HAP-NodeJS/commit/a4067085589753198e177786b151165824c19f99

**This is a breaking change**
Buffer.alloc/from are not supported in node 4. Not supporting node 4 is a bit harsh. Node 4 is still LTS and default package manager version in many OS distros and widely used. Is it really necessary?

See issue in telldus plugin:
https://github.com/jchnlemon/homebridge-telldus/issues/27

KhaosT:
Nodejs v7 plans to drop support for the original Buffer constructors. Because of that, I have to either migrate to "new Buffer(…)" or "Buffer.alloc"

Since Buffer constructors are marked as deprecated, I don't think it's a good idea to migrate to "new Buffer(…)" solution.

What's your proposed solution for this?

mifi:
Maybe something like this:
https://www.npmjs.com/package/buffer-shims

KhaosT:
I'm not sure... It feels like putting a lot of work on something just because people prefers to use a older version of software. It's not like nodejs v6 drops support for any hardware. If you really think it's worth the effort, feel free to send my a PR and I'm happy to merge it.

mifi:
True, but it breaks npm upgrade for a lot of people. I added a PR #324 where i replaced all Buffer.alloc and Buffer.from calls with buffer-shims. It's easy to go back on this in the future. Just find-replace reverse

KhaosT:
Thanks for the effort. v0.4.13 should work with earlier version of nodejs now.

PR #324, opened 2016-11-03
Use bufferShim for node 4 support

mifi:
Replace all Buffer.alloc and Buffer.from calls with buffer-shims
Also "npm install" has corrected the order of dependencies

My editor (Atom) has automatically removed indentation on empty lines too

I have tested on node4 and node6 and it solves the issue (#322)

### DIFF
--- a/lib/HAPServer.js
+++ b/lib/HAPServer.js
@@ -320,7 +320,7 @@ HAPServer.prototype._handlePairStepOne = function(request, response, session) {
   srp.genKey(32, function (error, key) {
 
     // create a new SRP server
-    var srpServer = new srp.Server(srpParams, bufferShim.from(salt), from("Pair-Setup"), bufferShim.from(this.accessoryInfo.pincode), key);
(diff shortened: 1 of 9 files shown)


lib/HAPServer.js
```diff
@@ -317,21 +318,21 @@ HAPServer.prototype._handlePairStepOne = function(request, response, session) {
   var srpParams = srp.params["3072"];
 
   srp.genKey(32, function (error, key) {
-    
+
     // create a new SRP server
-    var srpServer = new srp.Server(srpParams, Buffer.from(salt), Buffer.from("Pair-Setup"), Buffer.from(this.accessoryInfo.pincode), key);
+    var srpServer = new srp.Server(srpParams, bufferShim.from(salt), from("Pair-Setup"), bufferShim.from(this.accessoryInfo.pincode), key);
```
KhaosT: Please fix ``from("Pair-Setup")``

mifi: Wow nice catch. Don't know how that happened. I'll fix and take another look over the whole PR too.

mifi:
I reviewed the changes. Looks fine to me now.

KhaosT:
Looks good

Issue #331, opened 2016-11-13
Negative Temperature


Fafex:
Hello,

I've created a Temperature accessory for my home automation system. It should read current temperature from a txt file (that's a way I prefer). Every time the temperature goes below zero, every iOS app throws an error and refuses to show the temperature. I already changed minValue in both my accessory and in HomeKitTypes.js

Here's the part that reads temperature from my *_accessory.js

```
},{
    	cType: types.CURRENT_TEMPERATURE_CTYPE,
    	onRead: function(callback) {
    		console.log("Reading temperature outside");
    		fs = require('fs')
			fs.readFile('/home/username/hktmp.txt', 'utf8', function (err,data) {
  				if (err) {
    			        return console.log(err);
 					}
 				console.log(data);
 				var inttemp = parseFloat(data);
				console.log(inttemp)
  				callback(inttemp);
  			})
    	},
    	perms: ["pr","ev"],
		format: "float",
		initialValue: 24,
		supportEvents: false,
		supportBonjour: false,
		minValue: -100,
		maxValue: 100,
		manfDescription: "Temperature"
	},{
```

Console log, after parseFloat, throws out a correct value of negative temperature.

And the fragment I changed in HomeKitTypes.js

```
Characteristic.CurrentTemperature = function() {
  Characteristic.call(this, 'Current Temperature', '00000011-0000-1000-8000-0026BB765291');
  this.setProps({
    format: Characteristic.Formats.FLOAT,
    unit: Characteristic.Units.CELSIUS,
    maxValue: 100,
    minValue: -100,
    minStep: 0.1,
    perms: [Characteristic.Perms.READ, Characteristic.Perms.NOTIFY]
  });
  this.value = this.getDefaultValue();
};
```

Any idea what else can be going wrong here?

TheNoim:
For me it is simply working to change the min value to -100

Fafex:
Okay, so I've take a look at @TheNoim commit and my file was the same. So I've rewritten my _accessory.js file. It now reads temperature from a sensor through Domoticz API and parse the response with RegExp. It looks like this:

```
var Accessory = require('../').Accessory;
var Service = require('../').Service;
var Characteristic = require('../').Characteristic;
var uuid = require('../').uuid;
var request = require("request");

// here's a fake temperature sensor device that we'll expose to HomeKit
var FAKE_SENSOR = {
  getTemperature: function() { 
    console.log("Getting the current temperature!");
[...]

PR #333, opened 2016-11-13
Fix negative temperatures - issue #331

TheNoim:


KhaosT:
Those are metadata generated from HomeKit Accessory Simulator. If your accessory supports negative temperature, please override metadata using [``setProps``](https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/Characteristic.js#L108) instead of changing the universal metadata.

PR #346, opened 2016-11-24
Create simpleLight_accessory.js

OczkoSX:
In my opinion, better way of implementing new accessories (in this case simple light bulb with only on/off switch). Please consider adding.

### DIFF
--- a/accessories/simpleLight_accessory.js
+++ b/accessories/simpleLight_accessory.js
@@ -1,0 +1,57 @@
+var Accessory = require('../').Accessory;
+var Service = require('../').Service;
+var Characteristic = require('../').Characteristic;
+var uuid = require('../').uuid;
+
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks for spending time working on the new example and it looks good. However, I think the example suppose to show people how to interact with Characteristic, Service and Accessory. For this part, the new ``simpleLight_accessory.js`` is really similar to the existing ``Light_accessory.js`` and the only main difference is how you decided to group the information into the object. For now I'll just close the PR and if you really think it's necessary to have another similar example, let me know and I'll reconsider.

OczkoSX:
You're right, it was based on `Light_accessory.js`. This kind of grouping elements is easier to read in my opinion. Maybe I would add some characteristics and then put it instead of `Light_accessory.js`?

KhaosT:
@OczkoSX Yeah if you prefer, make update to that sounds better :)

PR #348, opened 2016-11-25
Create extendedLight_accessory.js

OczkoSX:
Added more characteristics than in Light_accessory.js, in my opinion code is much more cleaner and easier for further expanding.


KhaosT:
Sorry I think I misunderstood your description in #346. What I mean is for you to make changes directly onto ``Light_accessory.js`` instead of spin up another "light" accessory...

OczkoSX:
Ahh. Can't you just replace their content? People would know how to add hue, saturation and some other characteristics

KhaosT:
Okay I'll do that.

PR #350, opened 2016-11-26
Motion Sensor File

legotheboss:
Created a sample motion sensor file, would be a great resource to add to the accessories folder.


KhaosT:
Would you mind I ask what's the purpose of this...? It's not like this demonstrates how to inform HomeKit about value changes out side the HomeKit and it doesn't provide anything new compares to existing examples.

legotheboss:
Thought it might be a good resource for anyone who wants to create a motion sensor, but did not want to rewrite a lot of code...

KhaosT:
If you say so ^^ Personally I still think people should at least have basic understanding of what they are doing instead of just edit stuff they don't understand.

smith288:
I think these types of examples are great for beginner and people just wanting to template off it.

Maxmudjon:
how to add MQTT ?

PR #377, opened 2017-01-09
option to destroy the server, even when not started

neophob:


KhaosT:
Looks good, thanks :)

PR #385, opened 2017-01-12
Avahi advertiser

neophob:


KhaosT:
Can you elaborate a little bit more about this change and why is this needed?

neophob:
Sure, I want to use my existing mdns service (Avahi) - as it is already running and it doesn't make sense to have two mdns server running on one machine.

This means the mdns dependency is not absolutely needed anymore, that's why i moved the mdns dependency as optional dependency. 

Now if you define the env variable `AVAHI_TARGET_FILE` then you use HAP-NodeJS the internal mdns service will not be started but the service file for Avahi will be created / updated. I already use this and it works fine.

Hope that explains the use case. I will send more updates soon, including an updated readme. however other changes are not related to this PR, so I try to keep them separated.

KhaosT:
@neophob wait... the ``mdns`` dependency of this project is actually [this](https://github.com/agnat/node_mdns) which uses Avahi daemon on linux platform, it's not say, running two separate mdns services on the machine. Shouldn't this be the same as your file based solution?

neophob:
Well the `mdns` node package might be using some Avahi code, the point is that the `mdns` packages starts an embedded mdns server.

When I speak about Avahi I mean the Linux daemon Avahi.

So basically HAP-NodeJS can support

- mdns out of the box (existing)
- create an Avahi service file to use with Avahi daemons (this PR)

So this address two different use case's.

KhaosT:
It doesn't start an embedded mdns server... It uses avahi_client_new to connect to Avahi daemon and have it publish the service.

neophob:
@KhaosT you're right, I should have read the mdns readme. I'll check my use case again and will reopen an this ticket if needed. thanks

PR #388, opened 2017-01-14
Cleanup

neophob:
minor cleanup:
- move strict statement on top of classes
- replace usage of console.log with debug
- removed unused imports
- add removeAllBridgedAccessories function to remove all added accessoires 
- bump node-persistence
- minor update of the readme


lib/Accessory.js
```diff
@@ -11,9 +13,6 @@ var Advertiser = require('./Advertiser').Advertiser;
 var HAPServer = require('./HAPServer').HAPServer;
 var AccessoryInfo = require('./model/AccessoryInfo').AccessoryInfo;
 var IdentifierCache = require('./model/IdentifierCache').IdentifierCache;
-// var RelayServer = require("./util/relayserver").RelayServer;
```
KhaosT: Please don't remove this one, while it's not being publicly distributed, it's being used somewhere.

lib/gen/HomeKitTypes.js
```diff
@@ -640,6 +641,7 @@ Characteristic.CurrentTemperature = function() {
     format: Characteristic.Formats.FLOAT,
     unit: Characteristic.Units.CELSIUS,
     maxValue: 100,
+    //TODO see https://github.com/plavrenko/HAP-NodeJS/commit/53bc5635ed453135d6e48759cd987bae2a97c407
```
KhaosT: Why do you add this?

package.json
```diff
@@ -13,8 +13,8 @@
     "ed25519": "^0.0.4",
     "fast-srp-hap": "^1.0.0",
     "ip": "^1.1.3",
-    "mdns": "^2.3.2",
-    "node-persist": "^0.0.8"
+    "mdns": "^2.3.3",
+    "node-persist": "^0.0.11"
```
KhaosT: Maybe we should try and see if we can migrate to latest node-persist. It's fine to use 0.0.11 for this PR.

neophob: yeah, I guess it make sense to update node-persist in a separate PR

lib/gen/HomeKitTypes.js
neophob: ok will remove this, was more a todo for myself I forgot to add

lib/Accessory.js
neophob: ok, I'll re-add this. maybe control this with a env variable would make it easier to understand for the future

neophob:
@KhaosT updated the two issues you mentioned

KhaosT:
Thank You :)

PR #391, opened 2017-01-20
Fixed the fan speed

ddashwood:
When setting the fan speed for a "fake" fan, this fix ensures that we remember the fan speed so that Apple Home shows it correctly.


KhaosT:
Thanks :)

ddashwood:
No problem! I’m new to the Raspberry Pi, so fixing this was a nice gentle introduction!

PR #402, opened 2017-02-03
set homekit identification using env variables

neophob:


neophob:
any comments?

KhaosT:
Sorry forgot about this. Given HAP-NodeJS in never intended to be use as standalone software (it's more of a framework to build on top of it), I don't think it's necessary to allow people to change default information using env variable. Instead, people should change it in the proper way, setting those via characteristics.

neophob:
Thanks @KhaosT for the update, sounds plausible - for all others, see the `MotionSensor_accessory.js` example

PR #426, opened 2017-04-02
change curve25519-n to curve25519-n2 to support Windows

MystK:
curve25519-n isn't building on Windows 10. curve25519-n2 includes code to handle building on Windows 10. There are no other changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "dependencies": {
     "buffer-shims": "^1.0.0",
-    "curve25519-n": "^1.1.0",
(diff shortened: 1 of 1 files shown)


KhaosT:
Is this a public package? I can't find ``curve25519-n2`` in npmjs.com

MystK:
Yes, I just published it so that's probably why you can't find it in the search.

Linked here: https://www.npmjs.com/package/curve25519-n2

KhaosT:
Looks good.

PR #428, opened 2017-04-08
Update code dependency to use curve25519-n2

pdlove:
Completes the move from curve25519-n to curve25519-n2


KhaosT:
Oops nice catch 😅

PR #436, opened 2017-04-25
Update AccessoryLoader.js

OczkoSX:
Check if passed accessory is not empty, it was throwing an error on forEach at line 66 if so. To test it out, simply create an empty file called xxx_accessory.js.

### DIFF
--- a/lib/AccessoryLoader.js
+++ b/lib/AccessoryLoader.js
@@ -45,8 +45,11 @@ function loadDirectory(dir) {
 
   // now we need to coerce all accessory objects into instances of Accessory (some or all of them may
   // be object-literal JSON-style accessories)
-  return accessories.map(function(accessory) {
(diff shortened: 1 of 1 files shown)


lib/AccessoryLoader.js
```diff
@@ -45,8 +45,11 @@ function loadDirectory(dir) {
 
   // now we need to coerce all accessory objects into instances of Accessory (some or all of them may
   // be object-literal JSON-style accessories)
-  return accessories.map(function(accessory) {
-    return (accessory instanceof Accessory) ? accessory : parseAccessoryJSON(accessory);
+  return accessories.filter(function(accessory) {
```
KhaosT: `Array.filter` doesn't handle `parseAccessoryJSON()` properly.

OczkoSX:
My fault. I've got it working now. I will close this request and open new one.

PR #438, opened 2017-04-25
Create multipleLightAccessories_accfactory.js

OczkoSX:
Create multiple lights within one file with very easy and simple config


KhaosT:
This one will not be merged since I don't want to encourage people use hap-nodejs as a standalone program. Accessories under `/accessories` mainly serve as examples and in this case, having one light and multiple is the same.

OczkoSX:
So maybe show them how to use the accfactory? So they can do it on themselves. I was lucky that i found it in the files. That was the thing i tried to achieve with this example

KhaosT:
No, people should create their own js file to start the accessory instead of relying on Core/BridgedCore.

OczkoSX:
Okay, sure. I got it

PR #460, opened 2017-06-09
Updates for IOS 11 / XCode 9

pdlove:
There are many changes in the HomeKit Accessory Simulator metadata. Please review this list and let me know if anything needs to be manually put back:
Camera Related Items:
	Characteristic.SelectedStreamConfiguration renamed to Characteristic.SelectedRTPStreamConfiguration
	Service.CameraControl Removed
	Characteristic.DigitalZoom no longer attached to any service
	Characteristic.ImageMirroring no longer attached to any service
	Characteristic.ImageRotation no longer attached to any service
	Characteristic.NightVision no longer attached to any service
	Characteristic.OpticalZoom no longer attached to any service
	Characteristic.StreamingStatus is now writable
	
Utility Items:
	Service.Label renamed to Service.ServiceLabel
	Characteristic.LabelNamespace renamed to Characteristic.ServiceLabelNamespace
	Characteristic.LabelIndex renamed to Characteristic.ServiceLabelIndex
	Characteristic.AppMatchingIdentifier removed
	Characteristic.ProgrammableSwitchOutputState removed
	Characteristic.SoftwareRevision removed
	Service.StatefulProgrammableSwitch removed
	Service.AccessoryInformation now has Characteristic.FirmwareRevision as Required
	Service.AccessoryInformation no longer has Characteristic.SoftwareRevision as an associated Characteristic
	Service.AccessoryInformation no longer has Characteristic.AppMatchingIdentifier as an associated Characteristic

Other Changes:
	Characteristic.VOCDensity changed from FLOAT to UINT8
	Characteristic.ProgrammableSwitchEvent can now have any UINT8 value.
	Characteristic.ChargingState no longer has a maximum value of 2, but the Enumeration for it still only has up to 2 items.
	
Added Items:
	Characteristic.ColorTemperature

Manual Changes I've made:
	Associated Characteristic.ColorTemperature as an optional characteristic of the Light Bulb.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -103,12 +103,11 @@ Accessory.Categories = {
   CAMERA: 17,
   IP_CAMERA: 17, //Added to conform to HAP naming
   VIDEO_DOORBELL: 18,
-  AIR_PURIFIER: 19
-  //20+ is marked as reserved in the official protocol spec
-  //AIR_HEATER: 20,
-  //AIR_CONDITIONER: 21,
-  //AIR_HUMIDIFIER: 22,
-  //AIR_DEHUMIDIFIER: 23
+  AIR_PURIFIER: 19,
+  AIR_HEATER: 20, //Not in HAP Spec
+  AIR_CONDITIONER: 21, //Not in HAP Spec
+  AIR_HUMIDIFIER: 22, //Not in HAP Spec
+  AIR_DEHUMIDIFIER: 23 // Not in HAP Spec
 }
 
 Accessory.prototype._identificationRequest = function(paired, callback) {
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -94,18 +94,21 @@ Accessory.Categories = {
   THERMOSTAT: 9,
   SENSOR: 10,
   ALARM_SYSTEM: 11,
+  SECURITY_SYSTEM: 11, //Added to conform to HAP naming
   DOOR: 12,
   WINDOW: 13,
   WINDOW_COVERING: 14,
   PROGRAMMABLE_SWITCH: 15,
   RANGE_EXTENDER: 16,
   CAMERA: 17,
+  IP_CAMERA: 17, //Added to conform to HAP naming
   VIDEO_DOORBELL: 18,
-  AIR_PURIFIER: 19,
-  AIR_HEATER: 20,
-  AIR_CONDITIONER: 21,
-  AIR_HUMIDIFIER: 22,
-  AIR_DEHUMIDIFIER: 23
+  AIR_PURIFIER: 19
+  //20+ is marked as reserved in the official protocol spec
+  //AIR_HEATER: 20,
+  //AIR_CONDITIONER: 21,
+  //AIR_HUMIDIFIER: 22,
+  //AIR_DEHUMIDIFIER: 23
 }
 
 Accessory.prototype._identificationRequest = function(paired, callback) {
@@ -168,6 +171,34 @@ Accessory.prototype.addService = function(service) {
   }.bind(this));
 
   return service;
+}
+
+Accessory.prototype.setPrimaryService = function (service) {
+    //find this service in the services list
+    var targetServiceIndex;
+    for (var index in this.services) {
+        var existingService = this.services[index];
+
+        if (existingService === service) {
+            targetServiceIndex = index;
+            break;
+        }
+    }
+     
+    if (targetServiceIndex) {
+        //If the service is found, set isPrimaryService to false for everything.
+        for (var index in this.services)
+            this.services[index].isPrimaryService = false;
+
+        //Make this service the primary
+        existingService.isPrimaryService = true
+
+        if (!this.bridged) {
+            this._updateConfiguration();
+        } else {
+            this.emit('service-configurationChange', clone({ accessory: this, service: service }));
+        }
+    }
 }
 
 Accessory.prototype.removeService = function(service) {
--- a/lib/Characteristic.js
+++ b/lib/Characteristic.js
@@ -70,7 +70,9 @@ Characteristic.Formats = {
   UINT32: 'uint32',
   UINT64: 'uint64',
   DATA: 'data',
-  TLV8: 'tlv8'
+  TLV8: 'tlv8',
+  ARRAY: 'array', //Not in HAP Spec
+  DICTIONARY: 'dictionary' //Not in HAP Spec
 }
 
 // Known HomeKit unit types
@@ -92,7 +94,7 @@ Characteristic.Perms = {
   NOTIFY: 'ev', //Kept for backwards compatability
   EVENTS: 'ev', //Added to match HAP's terminology
   ADDITIONAL_AUTHORIZATION: 'aa',
-  TIMED_WRITE: 'tw',
+  TIMED_WRITE: 'tw', //Not currently supported by IP
   HIDDEN: 'hd'
 }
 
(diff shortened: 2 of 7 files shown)


lib/gen/HomeKitTypes.js
```diff
@@ -1478,9 +1439,6 @@ Characteristic.ProgrammableSwitchEvent = function() {
   Characteristic.call(this, 'Programmable Switch Event', '00000073-0000-1000-8000-0026BB765291');
   this.setProps({
     format: Characteristic.Formats.UINT8,
-    maxValue: 2,
```
KhaosT: This section is needed to support older iOS devices.

lib/gen/HomeKitTypes.js
```diff
@@ -6,47 +6,6 @@ var Characteristic = require('../Characteristic').Characteristic;
 var Service = require('../Service').Service;
 
 /**
- * Characteristic "Label Namespace"
- */
-
-Characteristic.LabelNamespace = function() {
```
KhaosT: To keep existing plugins working, I think we should create alias for these instead of just removing it.

KhaosT: Possibly move to another file to reduce the amount of manual work.

lib/gen/HomeKitTypes.js
```diff
@@ -1497,26 +1455,6 @@ Characteristic.ProgrammableSwitchEvent.DOUBLE_PRESS = 1;
 Characteristic.ProgrammableSwitchEvent.LONG_PRESS = 2;
 
 /**
- * Characteristic "Programmable Switch Output State"
- */
-
-Characteristic.ProgrammableSwitchOutputState = function() {
```
KhaosT: This one can probably be moved to deprecated file (to prevent existing plugin from crashing).

lib/gen/HomeKitTypes.js
```diff
@@ -1707,21 +1645,21 @@ Characteristic.SecuritySystemTargetState.NIGHT_ARM = 2;
 Characteristic.SecuritySystemTargetState.DISARM = 3;
 
 /**
- * Characteristic "Selected Stream Configuration"
+ * Characteristic "Selected RTP Stream Configuration"
  */
 
-Characteristic.SelectedStreamConfiguration = function() {
```
KhaosT: Guess alias for this too.

lib/gen/HomeKitTypes.js
```diff
@@ -1800,23 +1779,6 @@ Characteristic.SmokeDetected.SMOKE_NOT_DETECTED = 0;
 Characteristic.SmokeDetected.SMOKE_DETECTED = 1;
 
 /**
- * Characteristic "Software Revision"
- */
-
-Characteristic.SoftwareRevision = function() {
```
KhaosT: Move to deprecated probably.

ebaauw:
> Possibly move to another file to reduce the amount of manual work.

Makes sense.  It looks like `HomeKitTypes-Bridge.js` is already a manually edited file for deprecated (or no longer documented?) services and characteristics.  @KhaosT, should we reuse this file or do you want to create a new file for the manual stuff?

> To keep existing plugins working
> to prevent existing plugin from crashing

I assume you mean homebridge plugins?  This is a consequence of passing HAP-NodeJS to plugins through the homebridge API instead of having plugins `require` HAP-NodeJS themselves.  This is starting to look like a maintenance nightmare.  I think we can expect more frequent changes, now that Apple has documented HAP.

It might be something to reconsider for [homebridge 0.5](https://github.com/nfarina/homebridge/issues/1039).  I don't suppose homebridge could handle multiple plugins, each using a different version of HAP-NodeJS?

pdlove:
If the plugins are only using HAP-NodeJS for the Service/Characteristic mappings then the version shouldn't matter because it's just resolving to UUIDs and not really any of the functionality in HAP-NodeJS.

I don't think it's an issue to update the HomeKitTypes.js from HAS each release and put the differences into the HomeKitTypes-Bridge or another one called HomeKitTypes-Obsolete.js or HomeKitTypes-BeforeIOS11.js so that developers can see the supported types by IOS release.

The problem is that if Apple changes the constraints of an existing UUID, like they appear to have done with the ProgrammableSwitchEvent, then plugins could break on older versions of IOS.
A good example is that BatteryChargingState doesn't have a maxValue anymore so if they add another option in the Enumeration used by the characteristic then anything using that value will cause IOS versions before 11 to error.

I don't think it does, but I'll ask anyway.... does the HomeKit pairing process expose the IOS release? I know there's a protocol version but I don't think it changes when they make  these kind of potentially breaking changes on the Characteristics.

Now is definitely a good time to decide how to handle this.

ebaauw:
> The problem is that if Apple changes the constraints of an existing UUID, like they appear to have done with the ProgrammableSwitchEvent, then plugins could break on older versions of IOS.
A good example is that BatteryChargingState doesn't have a maxValue anymore so if they add another option in the Enumeration used by the characteristic then anything using that value will cause IOS versions before 11 to error.

See section 8.1.1 in the HAP spec: characteristic properties like minValue, maxValue, stepValue may be modified (per accessory).  A HomeKit app (or homebridge plugin) should not assume these are fixed, but check the actual values.  I'm more worried about changes like Volume becoming INT8 instead of FLOAT.

KhaosT:
When I say creating alias for supporting existing plugins, I mean for something like `Characteristic.LabelNamespace`, the actual version in the metadata plist is `Characteristic.ServiceLabelNamespace`. If we just make the change without alias, then existing plugin will throw exception with updated version since `Characteristic.LabelNamespace` no longer exists. It doesn't matter if the UUID is the same, it's the object referred by the name is gone.

HomeKit pairing doesn't expose iOS version so there is no way to determine that. However, HomeKit on iOS will follow the metadata provided by the accessory when the value is different from default. Take example of that charging state characteristic, to provide backward support for iOS, ideally we should always set metadata for min/max/step values so older iOS version will work correctly when accessory return a value that's larger than previous enum.

lib/Accessory.js
```diff
@@ -94,18 +94,21 @@ Accessory.Categories = {
...
+  IP_CAMERA: 17, //Added to conform to HAP naming
   VIDEO_DOORBELL: 18,
-  AIR_PURIFIER: 19,
-  AIR_HEATER: 20,
-  AIR_CONDITIONER: 21,
-  AIR_HUMIDIFIER: 22,
-  AIR_DEHUMIDIFIER: 23
+  AIR_PURIFIER: 19
+  //20+ is marked as reserved in the official protocol spec
+  //AIR_HEATER: 20,
```
KhaosT: Please try to leave these here for now as some people might already been using these.

lib/Characteristic.js
```diff
@@ -65,13 +65,11 @@ Characteristic.Formats = {
   INT: 'int',
   FLOAT: 'float',
   STRING: 'string',
-  ARRAY: 'array', // unconfirmed
```
KhaosT: Same for these, although i don't think anyone actually use these, let's mark it as deprecated.

lib/Characteristic.js
```diff
@@ -87,9 +85,14 @@ Characteristic.Units = {
...
-  WRITE: 'pw',
-  NOTIFY: 'ev',
+  READ: 'pr', //Kept for backwards compatability
+  PAIRED_READ: 'pr', //Added to match HAP's terminology
+  WRITE: 'pw', //Kept for backwards compatability
+  PAIRED_WRITE: 'pw', //Added to match HAP's terminology
+  NOTIFY: 'ev', //Kept for backwards compatability
+  EVENTS: 'ev', //Added to match HAP's terminology
+  ADDITIONAL_AUTHORIZATION: 'aa',
+  TIMED_WRITE: 'tw',
```
KhaosT: Timed write for IP is not supported for now... maybe mark that?

lib/Characteristic.js
```diff
@@ -156,7 +168,92 @@ Characteristic.prototype.getValue = function(callback, context, connectionID) {
...
+      maxValue_resolved=18446744073709551615;
+      isNumericType=true;
+      break;
+    case 'data':
+      var maxLength = this.props.maxDataLen;
+      if (maxLength===undefined) maxLength=2097152; //Default Max Length is 2097152.
+      if (newValue.length>maxLength) throw new Error('Characteristic ' + this.displayName + ' Failed Length Validation');
+      break;
+    case 'tlv8':
+      //Should we parse this to make sure the tlv8 is valid?
```
KhaosT: tlv8 and data share the same thing I believe.

lib/Characteristic.js
pdlove: I'll re-add the removed items with a note that they aren't supported according to the HAP spec.

lib/Characteristic.js
pdlove: I assumed data was a raw binary format and tlv8 is the structured data. My comment here is that the we should be able to parse a tlv8 response with a warning if an unspec'd datatype is used. I honestly don't think this kind of validation is worth the work but I wanted to point it out here.

PR #468, opened 2017-07-08
remove persisted files when accessory is destroyed

mschm:
Hi,
I want to propose this solution to remove all belongings of an accessory which is being destroyed.
Thank you!


mrose17:
+1 i think this is a very useful enhancement!

j4k3:
I've been complaining about this problem months ago, nice to see someone with experience in nodejs finally fixing it.

KhaosT:
Thanks :)

PR #483, opened 2017-08-16
Allow any characteristic to auto-add to a service when getting it.

pdlove:
This is intended to fix nfarina/homebridge#1433
The issue is that Apple removed an optional characteristic and so now, while the characteristic still exists, it's no longer in the optional list. This change would allow "getCharacteristic" to still auto-add it while displaying a warning so that the developer understands that the characteristic isn't in the optional list. If the developer uses "addCharacteristic" then the warning will not appear.
Is this a good way to handle this or should we just make the developer of that plugin fix the plugin to add?
I don't want to add the defunct characteristic as optional on the service because I want to make sure future projects don't see it as an option.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -381,10 +381,11 @@ Accessory.prototype._assignIDs = function(identifierCache) {
 
   for (var index in this.services) {
     var service = this.services[index];
-    if (this._isBridge)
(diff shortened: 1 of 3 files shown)


pdlove:
I've also put into this the fix for the IID issue which causes "Not Supported" on a single bridged device in IOS 11.

PR #485, opened 2017-08-16
Additional string validation added (fallback)

Samfox2:
Addresses an issue that causes a program crash if an empty string is submitted, e.g. manufacturer field in homebridge is empty. In that case string is defined as "NA"

### DIFF
--- a/lib/Characteristic.js
+++ b/lib/Characteristic.js
@@ -220,6 +220,7 @@ Characteristic.prototype.validateValue = function(newValue) {
       break;
     case 'string':
       var myString = newValue;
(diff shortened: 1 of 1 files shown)


Samfox2:
Should be done in plugin...

pdlove:
We assume that the plugin has at least basic sanity checks.
I've got on my todo list to clean up the checks for boolean. I'll consider this check as well in that because a null or undefined string would potentially cause bad data to be sent to IOS.

Samfox2:
Sounds great, at least a basic string validation would be nice :-)

PR #489, opened 2017-08-17
Backwards Compatibility Additions and Better DataType Management

pdlove:


snowdd1:
Hi @pdlove, does it also check for the indexed integer types, such as window contact which uses int but allows only 0 and 1, or Alarm device characteristics?

pdlove:
The check occurs here: https://github.com/KhaosT/HAP-NodeJS/blob/3901ac2de0cf07c3bb12667a10f41df96feb4f78/lib/Characteristic.js#L275
And import.js was updated to populate the validValues property as you can see here: https://github.com/KhaosT/HAP-NodeJS/blob/3901ac2de0cf07c3bb12667a10f41df96feb4f78/lib/gen/HomeKitTypes.js#L1674

snowdd1:
Cool. I'll need to scrap my own checking routine then, would be duplicated.

PR #490, opened 2017-08-20
Fixed bug with strange value from Siri

gijoecool:
* Fixed bug where Siri gives true/false in place of 1/0

### DIFF
--- a/lib/Characteristic.js
+++ b/lib/Characteristic.js
@@ -243,6 +243,8 @@ Characteristic.prototype.validateValue = function(newValue) {
     
   if (isNumericType) {
     if (isNaN(newValue)) return this.value; //This is not a number so we'll just pass out the last value.
+    if (newValue === false) return 0;
+    if (newValue === true) return 1;
     if ((!isNaN(this.props.maxValue))&&(this.props.maxValue!==null)) maxValue_resolved=this.props.maxValue;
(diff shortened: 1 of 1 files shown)


pdlove:
@gijoecool I agree this is needed. That said, isn't this only affecting  what is sent out from Homebridge to IOS? I'll accept this PR simply because so many plugins are sending true/false instead of 1/0 but I think there may be additional work to fix it if the issue is IOS originating the incorrect value. On homebridge-smartthings there aren't any hacks implemented to handle door locks although technically the code to set the lock isn't written properly.

gijoecool:
Maybe I misunderstand the code. When Siri sends the value as true/false and we are expecting a number, then it looks like HAP-NodeJS is not updating the value in homebridge. It is affecting the user because they ask Siri to lock/unlock the door and homebridge does nothing.

With that said, I think the bug is actually in iOS. I don't understand why Siri sends a different value than the Home app when locking/unlocking doors. At one point they were the same (probably pre 10.2). This might be fixed in future iOS releases. Maybe even 11... I haven't installed any betas or the GM.

pdlove:
@gijoecool That is correct. From what I understand there is a bug in the value IOS is sending. I haven't tested to see this first hand but based on the feedback I've heard from other developers... Apple messed it up.
Now that I've reviewed the code I see that the PUT call from IOS ends up using the Characteristic setValue call so this will work as intended. Thank you for submitting.

PR #491, opened 2017-08-21
expire unused ids optional

sokarovski:
If a bridge is published without any accessories and the accessories are added later the functionality to expire unused ids in the IdentifierCache is removing all the ID's and that's creating mess on the iOS later when the accessories are dynamically added. So i added functionality to disable the 'expire unusdes ids' also a check in the publish weather the accessory is bridge and the length of bridgedAccessories in the bridge is 0 then disable the 'expire unusdes ids'.


KhaosT:
Thanks :)

PR #512, opened 2017-10-11
Add Support for Alexa to control compatible HAP-NodeJS devices

NorthernMan54:
After 6 months of beta testing with users, I ready to submit this a pull request against the main version.   This changes adds an additional API to HAP-NodeJS, that support control from Alexa of compatible devices.  The implemented API, emulates enough of a Philips Hue Hub API to enable to Alexa to think that HAPNode-JS is a Hue hub.  Currently only LightBulb, Outlet, Switch, and Fan devices are exposed to Alexa.

At this time the feature can only be enabled via homebridge, I will send that pull that request as well.


KhaosT:
Hi there,
First I'd really like to thank you for spending time working on this, I'm glad it worked out for you.

However, I'm afraid I will not merge this to the library since the library itself is about acting as a simple layer for HAP and I don't think adding SSDP and fake Hue API is a reasonable thing to do in this library.

Please let me know if you think otherwise.

Thanks!

NorthernMan54:
I’m okay with not merging this in, do you have any thoughts on other integration patterns?  I was trying to do this without forcing a lot complexity onto the end users to enable the feature.  A homebridge plugin would work, but it would need access to all the accessories and be able to control them.  Are there any examples of that?

KhaosT:
Right now there isn't a proper way for homebridge plugin to get all accessories hosted by homebridge. Technically [this](https://github.com/nfarina/homebridge/issues/1039) should cover the need for a plugin to achieve this goal but right now there is no timeline for building that enhancement...

NorthernMan54:
I spent the evening looking at alternate approaches, and was looking at the event bus in HAP-NodeJS.  If the event bus from HAPServer was exposed to a homebridge plugin, it would expose all accessories and the ability to control.

My experience with the nodejs event framework is limited so if it is already exposed, sorry about the question.

KhaosT:
The HAPServer is not designed to be modified by plugin. The ideally solution would be have homebridge provide it's own abstraction on top of the HAP layer (accessory, service and characteristic), and then provide event hooks for plugin to access registered accessories.

PR #523, opened 2017-11-05
update IdentifierCache.js so that cache is persisted to disk only if …

simont77:
This PR is intended to avoid unneeded saving of IdentifierCache to disk.

On my RaspPI I noticed an unusual IO activity due to homebridge. At the end I discovered this is related to a periodic call of the "accessories" event by the clients, which then calls the AssignIDs function, eventually ending in any case with a call to IdentifierCache.Save.

This PR ensures that actual saving is performed only if really needed, by calculating the hash of the cache currently on disk and comparing it with that of the updated cache built by AssignIDs function. This compare method is the same already employed to determine if AccessoryInfo need to be saved. Note however two differences:
1 - Hash is not saved into the cache file (as opposed to what is done with AccessoryInfo), but calculated on load, stored in memory and updated on saving, in order not to break existing installations with a different cache file format
2 - Hashing and checking are done directly in the IdentifierCache class (in Load and Save method), and not by the caller (as done for AccessoryInfo), in order to be more transparent to future development.

### DIFF
--- a/lib/model/IdentifierCache.js
+++ b/lib/model/IdentifierCache.js
@@ -2,6 +2,7 @@
 
 var util = require('util');
 var storage = require('node-persist');
+var crypto = require('crypto');
 
 module.exports = {
   IdentifierCache: IdentifierCache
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks!

PR #529, opened 2017-12-03
Update minvalue of CurrentTemperature

simont77:
With this PR, negative values on CurrentTemperature characteristic are now allowed. This is useful for external temperature sensors, that with the current settings saturates to zero, making necessary an override in the plugin. I also tested the new negative boundary modifying the corresponding file in Homekit Accessory Simulator, and the resulting simulated accessory is correctly seen by Home.app and Eve.app, so there not seems to be any hard limit in HAP protocol.
I was reluctant to make this PR, since this file was originally autogenerated, but then I saw several manual commits, so I decided to try. I leave the decision to merge it or not to the admins.
Thanks

### DIFF
--- a/lib/gen/HomeKitTypes.js
+++ b/lib/gen/HomeKitTypes.js
@@ -680,7 +680,7 @@ Characteristic.CurrentTemperature = function() {
     format: Characteristic.Formats.FLOAT,
     unit: Characteristic.Units.CELSIUS,
     maxValue: 100,
-    minValue: 0,
+    minValue: -99, //manual update to support negative temperature on external sensors
     minStep: 0.1,
     perms: [Characteristic.Perms.READ, Characteristic.Perms.NOTIFY]
   });


ebaauw:
Might as well change it to -273.15.  I've been using this as `minValue` in homebridge-hue.  https://github.com/ebaauw/homebridge-hue/blob/36ebb2f43069e342416f063161ce3a61d0859d53/lib/HueSensor.js#L384

simont77:
True, but I limited it to -99 because it seems to me a reasonable limit for a real ambient temperature, but mainly because three digits negative numbers are not fully displayed in Home.app due to limited space.

NorthernMan54:
I believe this is a generated file, so I don't think they will accept this pull request.  To workaround this issue, I just set this in my homebridge plugins.

                this.temperatureService
                    .getCharacteristic(Characteristic.CurrentTemperature)
                    .setProps({
                        minValue: -100,
                        maxValue: 100
                    })
                    .on('get', this.getTemperature.bind(this));

This is an exert from my https://github.com/NorthernMan54/homebridge-dht/blob/master/index.js

simont77:
Actually I believe that starting from the commits for iOS 11 compatibility, this file is no more autogenerated. Since the issue is common to many plugin, it would be better to address it at an upper level,  instead of relying on each developer. Negative Celsius temperature are very common, and honestly I don’t understand why Homekit Accessory Simulator uses by default such a limited range for a float variable. Let’s see what KhaosT suggests.

Inviato da iPad

KhaosT:
Thanks for the PR. The problem here is that, unlike other manual changes to that file which is suppose to adopt hap-nodejs specific functionalities, this one is purely to address a special case where plugin developer forgot to handle in their plugin.

I have a few concerns:
- Most of HomeKit apps uses the min/max value to determine the slider steps when controlling the characteristic and setup triggers. By changing this to a really large value, people may have issue to set a proper value from the app.
- I worry this will set a bad precedent, people will try to create all kinds of workaround on the upper stream instead of fixing issues at the correct place.
- We don't have any knowledge on the underlying sensor, only plugin developer knows what's the expected range should be. Coming up with a random value is ... strange.

I can merge this if you guys really wants to, otherwise, I'd really suggest you talk to plugin developers that have this issue and let them fix it individually.

simont77:
Ok, I understand your points and will close the PR.
Do you have some idea about the right place, if any, for adding a note about that also for future plugin developers? The problem is that this issue seems to be often overlooked, and it is understandable, since it is so strange to have a min temp of zero degree. Actually the issue here is with Apple’s original implementation: -40 to +60 would have had the same span over a more natural range.

KhaosT:
@simont77 Thanks for the understanding ^^ I'm not exactly sure where is the best place to inform the developer that they need to override the value. Let me know if you have any suggestion :)

ebaauw:
> I have a few concerns:

> - Most of HomeKit apps uses the min/max value to determine the slider steps when controlling the characteristic and setup triggers. By changing this to a really large value, people may have issue to set a proper value from the app.

Yes, very annoying.  However, not applicable to CurrentTemperature, which is read-only.

> - I worry this will set a bad precedent, people will try to create all kinds of workaround on the upper stream instead of fixing issues at the correct place.

Fair point.

> - We don't have any knowledge on the underlying sensor, only plugin developer knows what's the expected range should be. Coming up with a random value is ... strange.

Hence my suggestion to use -273.15 (absolute zero).  Granted, still somewhat strange as no sensor will work at that temperature, but it’s not random :-).  The Hue motion sensor actually reports this as minimum value over ZigBee.  And 327.67 as maximum, as the value is an int16, representing centi-degrees Celsius.

PR #537, opened 2017-12-23
Minor Code / Dependencies Updates

jvmahon:
Minor code update to allow versions of node-persist >=1.0.0 to be used which altered the return syntax of GetItem. Updated other dependencies to recent versions.


package.json
```diff
@@ -10,16 +10,16 @@
   "dependencies": {
     "buffer-shims": "^1.0.0",
     "curve25519-n2": "^1.1.2",
-    "debug": "^2.2.0",
-    "ed25519": "^0.0.4",
+    "debug": "^3.1.0",
+    "ed25519": "git://github.com/jvmahon/ed25519",
```
KhaosT: Is there a reason why pin a git repo instead of the npm package?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.4.36",
+  "version": "0.5.0",
```
KhaosT: Please don't bump version number. It will be coordinated in release.

package.json
```diff
@@ -10,16 +10,16 @@
...
+    "debug": "^3.1.0",
+    "ed25519": "git://github.com/jvmahon/ed25519",
     "fast-srp-hap": "^1.0.1",
-    "ip": "^1.1.3",
-    "mdns": "^2.3.3",
-    "node-persist": "^0.0.11",
-    "decimal.js": "^7.2.3"
+    "ip": "^1.1.5",
+    "mdns": "^2.3.4",
+    "node-persist": "^2.1.0",
```
KhaosT: Does node-persist backwards compatible with data created in older version?

jvmahon:
Yes, though I’ll restore it for hap-nodejs to the npm package.

 

The ed25519 npm package won’t compile on Windows if you have the Open SSL library Version 1.1 or greater installed on Windows.

I’ve found the issue in ed25519 is a bindings library that changed. But I’ve now restored the hap-nodejs package to the npm version of ed25519 and I’ll raise the bindings library issue separately for ed25519

jvmahon:
Sorry about that. I’ve restored it to the original 4.36 numbering.

jvmahon:
Uncertain about that.

KhaosT:
Thanks, I'll try to test it this weekend :) Sorry for the delay.

jvmahon:
Thanks. As a practical matter, this change doesn't add new features or correct any bugs that I'm aware of - its more on the order of allowing the more updated node-persist and its features just in case others want to make improvements based on that. So its more of a "for consideration in your future development" kind of thing.

On a more relevant note, I've also been working on an update to the ed25519 module which does fix some issues. Specifically, the update gets rid of the many dozens of compiler warnings you get from the ed25519 module on Windows 64-bit (and possibly others?) during compile when installing hap-nodejs. The updated module simply uses the orlp/ed25519 updated ed25519 code base which has more explicit type casting. I have submitted this as a PR to the dozoe/ed25519 though its unclear if that repository is still getting updates. If it doesn't get incorported there, I'll put it on NPM.  I mention this because if you are doing further testing of the hap-nodejs, you may want to consider testing it with the updated ed25519. Its at https://github.com/jvmahon/ed25519 .  

Thanks for all your work on this software - its been a huge help to me and I'm sure to many others!  

As an FYI - I've also put a detailed Windows 64 bit install guide at http://github.com/jvmahon/homebridge in the file "HomeBridge on Windows 64 bit.docx".  I'm still working on that. Its currently in Word, but once its finished, I'll try to convert it to MD.  I mention this because it may also be helpful to those installing hap-nodejs without homebridge so, eventually, you might consider a link to it if you think it useful to users of your software who might not be also installing HomeBridge. I'll let you know once its finished if you want.

volschin:
In my understanding your ed25519 change breaks the compatibility with a lot of existing OpenSSL 1.0.x installations on Windows, also supported as STABLE until 2019. You need to implement a fix, that takes this into regard.

jvmahon:
volschin - Thanks for the info. The fix was actually intended, at least in part, to fix an incompatibility that arises when OpenSSL versions greater than 1.0.x are already installed on a Windows system by making the ed25519 code work with both pre- and post-1.0 versions of OpenSSL (OpenSSL versions > 1.0 didn't work on windows with the original ed25519 due to a renaming of OpenSSL's libeay32.lib to libcrypto.lib). I understand the explicit linking to the OpenSSL libeay32.lib in ed25519's bindings.gyp is no longer needed on Windows since nodejs includes OpenSSL -- thus, I commented out the explicit library link to libeay32.lib; as a result, the ed25519 code should rely on the OpenSSL linked into nodejs. It seems to work when I test it (at least on Windows 10 64 bit with nodejs 8). Can you point me to any info. discussing the incompatibility? Maybe it can be addressed while still allowing code that doesn't generate dozens of compiler warnings as was previously the case.

volschin:
Here is some general discussion about the topic 
https://github.com/nodejs/node/pull/13575

Using OpenSSL 1.1 in nodejs will make ed25519 accessible via internal crypto api and an additional module will no longer be needed. But this is open at the moment.

jvmahon:
volschin - Thanks. I appreciate your insights. I still don't really understand how my mods cause an OpenSSL compatibility break as you've suggested.  But maybe the concern can be narrowed. In my mods, there are two categories of changes - (1) the substitution of the orlp/ed25519 code for the original (this is really just about proper type casting and was the more "important" change in my view as it gets rid of dozens of warning errors when compiling on Win-64); and (2) removing the explicit libeay32.lib inclusion from the Windows-OS section in bindings.gyp and, instead, relying on the OpenSSL linked into nodejs.  Since change #1 is the more important (to me), I could revert the bindings.gyp changes (change #2) to its original form if that is where the OpenSSL break is that you are discussing (I'm still not sure there is a break, but i the interest of getting the #1 fixes potentially tested, #2 seems discardable).

Since this discussion is really about the ed25519 modification and no longer about the hap-nodejs, if you have further thoughts, I suggest continuing at my repository http://github.com/jvmahon/ed25519 so as not to clutter up this thread too much.

KhaosT:
Okay I did some tests and it looks like the changes in `node-persist` (1.0 and later) is not backward compatible with the original version we are using and they didn't provide a way to migrate (other than say bundle both version and manually set old data to new one)...
As a result of that, I'm afraid we have to leave `node-persist` in 0.0.11 for now...

jvmahon:
KhaosT - Since your testing shows the update to node-persist isn't backward compatible, I think it makes sense for me to close  this PR. I plan to close it tomorrow unless you suggest otherwise.

PR #539, opened 2017-12-30
Update value

NorthernMan54:
I have been moving my plugins away from using getValue to return the current accessory state, and switching to a event based model, and having the plugin push state via updateValue or setValue. ( I'm switching to this event based model, to improve the responsiveness of the Home app and reduce the lagging caused by the callouts to remote devices to return state. ) While this works really well, what is lacking is the ability to deal with accessories that are not responding for various reasons, as the updateValue and setValue don't accept a value of 'Error'.  And I'm not aware of any way to set a 'Not Responding' for an accessory either thru a characteristic or the HAP event interface,  except as a response to "GET /characteristics". 

1) Creates a characteristic level status 'this.status' - Value is null or the Error returned by the callback.

2) Changes updateValue and setValue to accept "Error" as a newValue, and update the status.

3) Changes getValue to include status as part of the response when there are no listeners to the get event

To use the feature, have the value updated by updateValue or setValue be an instance of Error.  During the next get characteristics request, the accessory will display 'Not responding'.  When the device starts responding again, sending a valid value, that is not an instance of 'Error' will clear the 'Not responding' message after the next get characteristics.


pdlove:
I must say I really like this approach. I think that the other thing that should be done, if it hasn't already, is to allow a getValue function to pass an Error status back to the callback.

NorthernMan54:
For the getValue callback, I’m also picking up the error returned as part of the callback, and storing that as the status.  So I’m think we are covered.  Do you have any other thoughts?

ebaauw:
+1 for liking this approach.  For my homebridge plugins, I've been using `setValue()` and `updateValue()` exclusively and stopped provided a listener for `get`, once I realised (many years ago) that HAPNode-JS returns the cached value anyways.  I've been thinking of a way to indicate unreachable accessories, and this seems like it would work.  I suppose each application/plugin could implement this themselves (by providing a listener for `get`), but it makes sense to provide this functionality out-of-the-box by HAP-NodeJS

Some thoughts:
- Do we need a separate `this.status` to hold the _not reachable_ condition?  Could we not use `this.value === null` instead?  Or are there any characteristic formats where `null` would be a valid value?
- Could we initialise the characteristics as _not reachable_ instead of having value `this.getDefaultValue()`?

pdlove:
I originally thought we did need a separate status because HAPServer.Status has multiple possible values but then I realized that this code isn't allowing the passing of the HAPServer.Status flag but is simply setting null. This method only allows passing SERVICE_COMMUNICATION_FAILURE but I'd like to be able to pass OPERATION_TIMED_OUT or RESOURCE_BUSY for "more correct" responses. I have no idea if these affect anything in IOS other than allowing Siri to say something a bit different. Maybe a nullable field at the end of updateValue and setValue that allows setting the specific Status and the Error method you're using would set that to SERVICE_COMMUNICATION_FAILURE. The only gotcha is that a non-error value and a null status parameter should then set this.status to SUCCESS (0)

I wouldn't think the characteristics should be initialized as not reachable because some are added with the intention of just leaving them at default. I think Homebridge, however, should set all characteristics as not reachable until their plugin is loaded, but that's a whole other phase.

NorthernMan54:
That is an interesting thought.   Taking this a further, what about just defining the err value to an 'Error' object with the message being the particular HAPServer.Status you want to return?

So a callback of an error condition would be something like 
callback( new Error( HAPServer.Status.SERVICE_COMMUNICATION_FAILURE ), null )

And adjust this code in HAPServer to look at the Error object, and if the message is a valid HAPServer.Status send that as the status, if not just send SERVICE_COMMUNICATION_FAILURE.

`if (!characteristics && !err)
                err = new Error("characteristics not supplied by the get-characteristics event callback");

            if (err) {
                debug("[%s] Error getting characteristics: %s", this.accessoryInfo.username, err.stack);

                // rewrite characteristics array to include error status for each characteristic requested
                characteristics = [];
                for (var i in data) {
                    characteristics.push({
                        aid: data[i].aid,
                        iid: data[i].iid,
                        status: HAPServer.Status.SERVICE_COMMUNICATION_FAILURE
                    });
                }
            }`

My thinking is that this approach would not break any existing accessories, and give a richer experience for those who want the functionality.

pdlove:
That works for me. I'm not very familiar with the Error object and a quick google didn't show that as something that could be done but it was my first thought. 👍

NorthernMan54:
Okay, let me make those changes to HAPServer and I will re-submit this pull request.

PS I was looking at the examples on this page for inspiration

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error](url)

KhaosT:
Is this PR ready to test?

grover:
Looking forward to be able to use this too.

NorthernMan54:
This pull request is not ready yet, I need to add an additional feature then it should be good to go.

NorthernMan54:
@pdlove @ebaauw I still need to do more regression testing around odd ball data conditions but this is conceptually complete based on everyone's input.  The only part of this I'm not thrilled with is validating that the status returned is a valid HAPServer.Status.   I used a for loop to iterate thru the HAPServer.Status object.   Any further thoughts before I wrap this up?

In my plugin, I used this a the value to return updateValue when an error occurred.

```
value = new Error( HAPServer.Status.SERVICE_COMMUNICATION_FAILURE );

this.getService(Service.TemperatureSensor).getCharacteristic(Characteristic.CurrentTemperature).updateValue(value);
```

For plugin developers, will need to determine an easy method to access the HAPServer.Status object to avoid everyone having to include a copy.

NorthernMan54:
@pdlove @ebaauw @KhaosT I have completed my testing on the final version and this is ready to go.

To use the feature, have the value updated by updateValue or setValue be an instance of Error. During the next get characteristics request, the accessory will display 'Not responding'. When the device starts responding again, sending a valid value, that is not an instance of 'Error' will clear the 'Not responding' message after the next get characteristics.   If you want to set a particular HAP return code, have the error be the HAP return code.

ie `new Error( HAPServer.Status.SERVICE_COMMUNICATION_FAILURE )`

or

`this.getService(Service.TemperatureSensor).getCharacteristic(Characteristic.CurrentTemperature).updateValue(new Error( HAPServer.Status.SERVICE_COMMUNICATION_FAILURE ));`

To gain access to the HAPServer Status codes, add these 2 HAPServer lines to the start of your plugin.   Please note that this is optional, and the feature will default to the status code HAPServer.Status.SERVICE_COMMUNICATION_FAILURE if you use a text or other error message.  So it should not break the existing behaviour of any plugins.

```
var Accessory, Service, Characteristic, UUIDGen, **HAPServer**;

module.exports = function(homebridge) {

  Accessory = homebridge.platformAccessory;
  Service = homebridge.hap.Service;
  Characteristic = homebridge.hap.Characteristic;
  UUIDGen = homebridge.hap.uuid;
  **HAPServer = homebridge.hap.HAPServer**;

  homebridge.registerPlatform("homebridge-ecoplugs", "EcoPlug", EcoPlugPlatform);
}

```

lib/Accessory.js
```diff
@@ -757,7 +757,7 @@ Accessory.prototype._handleGetCharacteristics = function(data, events, callback,
           aid: aid,
           iid: iid
         };
-        response[statusKey] = HAPServer.Status.SERVICE_COMMUNICATION_FAILURE; // generic error status
+        response[statusKey] = hapStatus(err);
```
grover: I'm a little bit concerned that line 755 uses %s for err.message, which in your examples is now a numeric value. I know JS will likely coerce the value here, but still I think this warants a change to %o there.

lib/Accessory.js
```diff
@@ -998,3 +998,18 @@ Accessory.prototype._generateSetupID = function() {
 
   return setupID;
 }
+
+function hapStatus(err) {
+
+  // Validate that the message is a valid HAPServer.Status
+  var value = 0;  // default if not found or
```
grover: Just leave it undefined here... Zero might also be something that could be in HAPServer.Status.

lib/Accessory.js
```diff
@@ -998,3 +998,18 @@ Accessory.prototype._generateSetupID = function() {
...
+
+  for( const k in HAPServer.Status ) {
+    if (HAPServer.Status[k] == err.message)
+      value = err.message;
+  }
+
+  if ( value == 0 )
```
grover: And check for undefined here.

grover:
Beyond those three comments I think this is great to finally have. Thank you!

lib/Accessory.js
NorthernMan54: @grover With the value = 0, I'm trying to avoid the valid "HAPServer.Status.SUCCESS: 0".   My thought was there might be a case with an existing plugin returning an err.message of 0 already.  And needed to ensure that it is still returned as an error.   My though is that this and 1012 should stay.

lib/Accessory.js
NorthernMan54: @grover At line 1014, I'm forcing to int

return(parseInt(value));

NorthernMan54:
I left comments on your three comments.  I feel the first was already covered, and what you are proposing for the other 2 breaks what I was trying to achieve.

index.js
```diff
@@ -20,7 +21,8 @@ module.exports = {
   Characteristic: Characteristic,
   uuid: uuid,
   AccessoryLoader: AccessoryLoader,
-  StreamController: StreamController
+  StreamController: StreamController,
+  HAPServer: HAPServer
```
KhaosT: Is there a reason why you are trying to expose HAPServer as part of the library? That class is not intended to be used separately...

NorthernMan54: @KhaosT I’m exposing HAPServer, to give the plugin developer access to HAPServer.status object, and allow them to optionally select a particular HAPServer.status to return. I chose to expose it this way, so that there is only I copy of it in the codebase incase in future release more status codes are added or the values change.  The ability to return a particular HAPServer status was requested by @pdlove and @ebaauw.

Returning a HAPServer.status is an optional feature for backwards compatibility.  If you use an error object containing any other value, it returns as a status the existing value, HAPServer.status.SERVICE_COMMUNICATION_FAILURE.

NorthernMan54:
@KhaosT Any more comments on this pull request?

lib/Accessory.js
KhaosT: Minor detail, we probably want a break at here so if there is a match, we don't always go over the entire Status array.

KhaosT:
@NorthernMan54 just one minor comment and then we are good to go :) Thanks!

lib/Accessory.js
```diff
@@ -998,3 +998,18 @@ Accessory.prototype._generateSetupID = function() {
...
 }
+
+function hapStatus(err) {
+
+  // Validate that the message is a valid HAPServer.Status
+  var value = 0;  // default if not found or
+
+  for( const k in HAPServer.Status ) {
+    if (HAPServer.Status[k] == err.message)
+      value = err.message;
```
KhaosT: Minor detail, we probably want a break at here so if there is a match, we don't always go over the entire Status array.

KhaosT:
It looks like last commit reverted a lot of stuff in the master right now 😅

NorthernMan54:
My baseline is out of date, let me redo

NorthernMan54:
@KhaosT I just pooched my GitHub fork, so this is going to take a day or 2.  Will let you know

KhaosT:
@NorthernMan54 Sorry about that, I should check it out earlier... Have you tried git rebase?

NorthernMan54:
Tried, but I’m maintaining my Alexa Fork of HAP-NodeJS as well, so things got complicated quickly.  I’m thinking to just create a second GitHub account, and start again.  As the change was tiny, rekeying isn’t an issue for me.

NorthernMan54:
Replacing with pull request 566

Issue #540, opened 2018-01-04
How to detect if a characteristic is subscribed?


grover:
Hi,

I'm using HAP-NodeJS through homebridge and am not sure where to file this issue... Anyways, I'm the author of [homebridge-ranger](https://github.com/grover/homebridge-ranger) - a range extender for HomeKit Bluetooth Accessories.

One of the issues I currently face is that some bluetooth accessories are very chatty if I unconditionally  subscribe all characteristics.  This in turn causes BLE connections to stay open, longer than necessary and in some instances the BLE accessory will not stop with the notifications.

The proper solution would be to know, for each characteristic, when and if a controller has subscribed it on the IP side.

Is there a solution for this today that I couldn't see from the code? Is this something that is desired from a contribution point of view? I can't imagine that I'm the only one that needs this when interfacing with third party devices/services.

Thanks!

grover:
Thanks for merging the pull requests.

Out of curiosity: how is the release process between this and homebridge? Are you taking care of it or do I need to create a separate issue in @nfarina‘s homebridge?

KhaosT:
I already changed dependency version on homebridge.

grover:
Just verified. Homebridge 0.4.35 contains the changes.

Thank you!

PR #542, opened 2018-01-05
Provide subscription status to characteristics

grover:
This pull request provides characteristics with subscription status knowledge. This is helpful to interact with external devices that are battery operated, but do provide on demand notifications of status changes. Especially helpful to implement the Bluetooth range extender. See #540.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -591,10 +591,10 @@ Accessory.prototype.publish = function(info, allowInsecureRequest) {
   this._server.on('accessories', this._handleAccessories.bind(this));
   this._server.on('get-characteristics', this._handleGetCharacteristics.bind(this));
   this._server.on('set-characteristics', this._handleSetCharacteristics.bind(this));
+  this._server.on('session-close', this._handleSessionClose.bind(this));
 
   if (this.cameraSource) {
       this._server.on('request-resource', this._handleResource.bind(this));
-      this._server.on('session-close', this._handleSessionClose.bind(this));
   }
 
   var targetPort = info.port || 0;
@@ -920,10 +920,31 @@ Accessory.prototype._handleResource = function(data, callback) {
   callback('resource not found');
 }
 
-Accessory.prototype._handleSessionClose = function(sessionID) {
+Accessory.prototype._handleSessionClose = function(sessionID, events) {
   if (this.cameraSource && this.cameraSource.handleCloseConnection) {
       this.cameraSource.handleCloseConnection(sessionID);
+  }
+
+  this._unsubscribeAllEvents(events);
+}
+
+Accessory.prototype._unsubscribeAllEvents = function (events) {
+  for (var key in events) {
+    if (key.indexOf('.') !== -1) {
+      try {
+        var id = key.split('.');
+        var aid = Number.parseInt(id[0]);
+        var iid = Number.parseInt(id[1]);
+
+        var characteristic = this.findCharacteristic(aid, iid);
+        if (characteristic) {
+          characteristic.unsubscribe();
+        }
+      }
+      catch (e) {
+      }
     }
+  }
 }
 
 // Called internally above when a change was detected in one of our hosted Characteristics somewhere in our hierarchy.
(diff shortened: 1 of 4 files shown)


grover:
My bad, will undo the formatting changes.

grover:
Cleaned up now.

lib/Characteristic.js
```diff
@@ -126,6 +128,21 @@ Characteristic.prototype.setProps = function(props) {
   return this;
 }
 
+
+Characteristic.prototype.subscribe = function() {
+  if (this.subscriptions++ === 0) {
```
KhaosT: Can we do
```js
if (this.subscriptions === 0) {
  this.emit('subscribe');
}
this.subscriptions++;
```
for improved readability?

lib/Characteristic.js
```diff
@@ -126,6 +128,21 @@ Characteristic.prototype.setProps = function(props) {
...
+    this.emit('subscribe');
+  }
+}
+
+Characteristic.prototype.unsubscribe = function() {
+  var wasOne = this.subscriptions === 1;
+  this.subscriptions = Math.max(this.subscriptions--, 0);
```
KhaosT: Same for this.

grover:
Sure can do.

lib/Accessory.js
```diff
@@ -916,10 +920,31 @@ Accessory.prototype._handleResource = function(data, callback) {
...
+  for (var key in events) {
+    if (key.indexOf('.') !== -1) {
+      try {
+        var id = key.split('.');
+        var aid = Number.parseInt(id[0]);
+        var iid = Number.parseInt(id[1]);
+
+        var characteristic = this.findCharacteristic(aid, iid);
+        if (characteristic) {
+          characteristic.unsubscribe();
```
KhaosT: Wouldn't this logic cause we over-unsubscribe for some characteristics (especially if there are multiple iOS devices connected to HAP server)? Say client didn't subscribe the characteristic at all, and then disconnects, we still invoke unsubscribe which will decrement subscriptions on the characteristic object...

grover: By my understanding of the code the events object is tied to the EventedHTTPConnection and this only unsubscribes those of that specific connection that’s gone away.

grover: See lines 842 and 845 of this file and line 310 in eventedhttp.js

KhaosT: @grover ah right, it's going through the events array. In that case would you mind to rename `_unsubscribeAllEvents` to `_unsubscribeEvents` and then we are good to go :)

grover: Will do.

lib/Accessory.js
```diff
@@ -841,10 +841,14 @@ Accessory.prototype._handleSetCharacteristics = function(data, events, callback,
       // the request.
       var eventName = aid + '.' + iid;
 
-      if (ev)
+      if (ev) {
```
grover: I think I just stumbled over another iOS weirdness: it issues multiple PUTs with ev: false. That does cause an over-unsubscription. I’ll add more checks here.

KhaosT:
Thank You :)

Krocko:
After this commits i have much problems with Homebridge (0.4.35)
After Homebridge is running for a while, my dash buttons don‘t work anymore, my sensors don’t change their states in the Home App (but in the Homebridge log) und my automations don‘t work anymore.

grover:
I’m saw this in my setup this morning too. I’m not sure where this comes from as it seems that it only happens after several hours. After a homebridge restart everything works as expected again until something happens...

Krocko:
Can we simply revert this changes for now?

grover:
Why? Just downgrade back to 0.4.34. I currently don’t see how the issue is connected. Yes they are related to events, but technically there’s no significant change that would explain this behavior IMO. I’m running with extended tracing today and already have a couple of other folks that are sending me their logs. I’m still hopeful to have the necessary understanding to make a more informed decision.

grover:
See https://github.com/KhaosT/HAP-NodeJS/pull/546 - I've added more logging to understand the issue. @Krocko - are you interested to run the version and provide me the logs after it happened?

PR #544, opened 2018-01-08
Unsubscribed events are deleted and not false

grover:
You were too quick :) 👍


KhaosT:
Oops... So is this good to go now?

grover:
I think so...

| ev | events[id] | sub? | unsub? |
|---|---|---|---|
| true | true | false | false |
| true | false | true | false |
| false | true | false | true |
| false | false | false | false |

Conditions passed those tests. Not sure if you see another one?

KhaosT:
Yeah I think that's all cases.

PR #551, opened 2018-01-24
Improve the new services with minValue, maxValue and validValues

renssies:
The new services for faucets, irrigation, and shower systems have finally been added. But they are missing `minValue`, `maxValue` and `validValues`. 

This PR adds them so that the new services are more in line with the others.


KhaosT:
Thank you :)

PR #558, opened 2018-02-08
Changes SSRC to be random per prepareStream, not always 1

aaronpearce:
This fixes the issue where you could not connect to multiple cameras at once when live.

SSRC should be a random 32 bit integer not always set to 1 for HomeKit to correctly be able to differentiate which stream is for which camera.

I have made it random on both audio and video streams, and have modified the ffmpeg call to correctly use this ssrc.

### DIFF
--- a/lib/Camera.js
+++ b/lib/Camera.js
@@ -99,9 +99,12 @@ Camera.prototype.prepareStream = function(request, callback) {
     let srtp_key = videoInfo["srtp_key"];
     let srtp_salt = videoInfo["srtp_salt"];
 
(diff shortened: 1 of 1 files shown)


KhaosT:
BTW I checked the RTP RFC and ffmpeg, it appears SSRC is expected to be an `Int32` instead of an `UInt32`. Using `UInt32` at here with ffmpeg may lead to unexpected behaviors.

aaronpearce:
Oh, missed that fact. Sorry everyone.

PR #565, opened 2018-03-08
Add libraries in readme

sprut666666:


simont77:
As already explained, I do not agree in suggesting to new developer a new custom UUID “standard” that is not standards at all. Maybe we should discuss a bit on that on slack.

PR #574, opened 2018-03-30
Find available tcp port before publishing accessory.

nikolamin:
Finds available TCP Port to publish the accessory.

### DIFF
--- a/Core.js
+++ b/Core.js
@@ -6,11 +6,23 @@ var accessoryLoader = require('./lib/AccessoryLoader');
 
 console.log("HAP-NodeJS starting...");
 
+// Checks is a tcp port available. 
+var isPortTaken = function(port, fn) {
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks for the PR but this feature is already built in underlying Accessory implementation 😅 You can publish an accessory without providing a port and it will figure out which port to use.

nikolamin:
@KhaosT Core.js is using targetPort++ to generate different port for each accessory, starting from 51826. In case port 51826 is busy, the accessory publishing will fail/crash.

KhaosT:
I understand that. The goal of `Core.js` is to show how to create an accessory. It was written before the server gain the ability to auto select the port. If you want, you can change the diff to publish the accessory without a port, that should achieve the same goal while not introducing duplicated logic.

PR #580, opened 2018-04-16
added accessory.unpublish()

### LINKED ISSUE #579: implement something like "accessory.unpublish()" or "accessory.isPublished()"
Hi,

I recently forked the node-red implementation of HAP-NodeJS and ran across an issue.
The current implementation was buggy when redeploying node-red because node state is supposed to be "reset" after a new deploy.

Unfortunately, there is no way to either determine if an accessory is currently being published or to "un-publish" it.

This leads to the problem that my new implementation tries to re-publish a service after redeploying a node-red flow which then crashes node-red:

```
16 Apr 21:18:15 - [red] Uncaught Exception:
16 Apr 21:18:15 - Error: dns service error: name conflict
    at Error (native)
    at Advertisement.on_service_registered (/usr/lib/node_modules/@oliverrahner/node-red-contrib-homekit-bridged/node_modules/mdns/lib/advertisement.js:42:24)
    at SocketWatcher.MDNSService.self.watcher.callback (/usr/lib/node_modules/@oliverrahner/node-red-contrib-homekit-bridged/node_modules/mdns/lib/mdns_service.js:18:40)
```

EDIT:
Just tried a 'proof-of-concept' using the option to un-publish the server.
This works:

```
   this.on('close', function (removed, done) {
      if (removed) {
        // This node has been deleted
        bridge.destroy()
      } else {
        // This node is being restarted
        bridge._server.stop()
        bridge._server = undefined
        bridge._advertiser.stopAdvertising()
        bridge._advertiser = undefined
        bridge = null
      }
      done()
    })
```

After that, the bridge seems to be properly cleaned up so that my re-publishing works.

oliverrahner:
Closes #579

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -613,14 +613,7 @@ Accessory.prototype.publish = function(info, allowInsecureRequest) {
  * Trying to invoke publish() on the object will result undefined behavior
  */
 Accessory.prototype.destroy = function() {
-  if (this._server) {
-    this._server.stop();
-    this._server = undefined;
-  }
-  if (this._advertiser) {
-    this._advertiser.stopAdvertising();
-    this._advertiser = undefined;
-  }
+  this.unpublish();
   if (this._accessoryInfo) {
       this._accessoryInfo.remove();
       this._accessoryInfo = undefined;
@@ -628,6 +621,17 @@ Accessory.prototype.destroy = function() {
   if (this._identifierCache) {
       this._identifierCache.remove();
       this._identifierCache = undefined;
+  }
+}
+
+Accessory.prototype.unpublish = function() {
+  if (this._server) {
+    this._server.stop();
+    this._server = undefined;
+  }
+  if (this._advertiser) {
+    this._advertiser.stopAdvertising();
+    this._advertiser = undefined;
   }
 }
 


KhaosT:
The concern here is that `this._server` this not completely stateless... I'm worried that if you call `unpublish()` and then publish the accessory again something will mess up.

oliverrahner:
Ok, I feared so... I think I can see that behavior in my own usage of this solution.

Any ideas how we could solve the issue I mentioned in #579 in another way?

KhaosT:
@oliverrahner Why can't you just `destroy` the old accessory object, create a new one and publish that?

KhaosT:
Okay the `unpublish` is here for a different reason... I can't believe I messed it up initially 😅

But still, for your situation do you intend to keep pairing information when publish another accessory?

oliverrahner:
Yes, I need to. 
node-red doesn't completely unload my module but intends to have some kind of "restart" for each node, which is the equivalent to the accesory level.

KhaosT:
`unpublish` is there now.

oliverrahner:
Thanks for the info! I'll try and see if that fixes all of my need!

PR #588, opened 2018-05-22
Sprinkler - Completed timer functionality & optimised logging

AndreasPrang:
The HomeKit timer functionality was not really implemented. Now it works.


AndreasPrang:
Fixed status updates.

KhaosT:
Thanks for contributing. However, the goal of those accessory examples is to show how to use the services, instead of making them simulate everything... Adding timer support introduces additional complexity that makes the control flow hard to follow as a basic example.

If you believe otherwise, feel free to reopen the PR.

PR #594, opened 2018-06-21
Add setupURI to the TypeScript def for Accessory

RLovelett:
The method was added in https://github.com/KhaosT/HAP-NodeJS/commit/c62294b3c5400157164a10a1e96a41259c3fc23b but was not added to the interface.


KhaosT:
Thank you :)

PR #595, opened 2018-06-21
catch oldValue in setValue directly at the beginning

Apollon77:
Catch oldValue in setValue directly at the beginning to make sure cha…nge events are triggered as soon as value change.

Reason: I found some Homebridge Plugins that break the whole "sending change events" process by setting the value itself in the set-handler. The result is that old and new value are the the same and no change event is triggered at all.

Example: https://github.com/YinHangCode/homebridge-mi-aqara/blob/master/parser/GatewayParser.js#L267

This change would fix this and really compare oldValue from timepoint of the setValue call with the new value

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,6 +30,10 @@ declare namespace HAPNodeJS {
         iid: string;
         characteristics: Characteristic[];
         optionalCharacteristics: Characteristic[];
+        
+        isHiddenService: boolean;
+        isPrimaryService: boolean;
+        linkedServices: Service[];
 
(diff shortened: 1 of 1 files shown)


Apollon77:
@KhaosT Do you have any opinion on this? :-)

Apollon77:
PS: in fact it is opnly the change in  lib/Characteristic.js ... if you would accept it I can recreate the PR based on most current HAP version

KhaosT:
Yeah looks good to me. Sorry about the long delay. Can you rebase from current master so I can merge this one? Thanks 🙏

Apollon77:
see #687

PR #631, opened 2018-12-15
Fix types of services exposed by Service

JosephDuffy:
I'm using this project is [one of my projects](https://github.com/JosephDuffy/homebridge-pc-volume) and found that the type that each of the services on `Service` have appears to be wrong.

Each predefined service property, e.g. [`Speaker`](https://github.com/KhaosT/HAP-NodeJS/blob/9eaea6df40811ccc71664a1ab0c13736e759dac7/lib/gen/HomeKitTypes.js#L3365), is exposed as a `Service`, which has a constructor with the signature `(displayName: string, UUID: string, subtype: string): Service`, but I believe they actually have the signature `(displayName: string, subtype: string): Service`, because the UUID is predefined.

I have added a new type – `PredefinedService` – which has a constructor with the signature `(displayName: string, subtype: string): Service`, and updated the `index.d.ts` file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,44 +42,48 @@ declare namespace HAPNodeJS {
 
         toHAP(opt: any): JSON;
 
-        AccessoryInformation: Service;
-        AirPurifier: Service;
-        AirQualitySensor: Service;
-        BatteryService: Service;
-        CameraControl: Service;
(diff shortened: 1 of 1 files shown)


KhaosT:
Thanks 🙏Sorry for the long delay...

PR #639, opened 2019-01-25
Update TV_accessory.js

lboue:
More inputs


KhaosT:
Thanks for contributing. The goal of this file is to provide a basic example of how to interact with the TV service and input sources. It's not intended to say, support everything that's possible. Because of that, this PR will not be merged.

PR #643, opened 2019-02-09
In insecure mode, force use of PIN for all operations

NorthernMan54:
As I have been widely using Insecure mode with homebridge-alexa, I found that /accessories and GET /characteristics can be accessed without supplying a PIN in insecure mode.  This pull request forces all insecure mode clients to supply the PIN.

For large scale projects potentially impacted by this change, I'm aware of my homebridge-alexa and @oznu and homebridge-config-ui-x.  So before merging, would need at a minimum Oznu to comment.

### DIFF
--- a/lib/HAPServer.js
+++ b/lib/HAPServer.js
@@ -850,6 +850,15 @@ HAPServer.prototype._handleAccessories = function(request, response, session, ev
     return;
   }
 
+  if (!session.encryption) {
+    if (!request.headers || (request.headers && request.headers["authorization"] !== this.accessoryInfo.pincode)) {
+      response.writeHead(401, {"Content-Type": "application/hap+json"});
+      response.end(JSON.stringify({status:HAPServer.Status.INSUFFICIENT_PRIVILEGES}));
+
+      return;
+    }
+  }
+
   // call out to listeners to retrieve the latest accessories JSON
   this.emit('accessories', once(function(err, accessories) {
 
@@ -901,6 +910,15 @@ HAPServer.prototype._handleCharacteristics = function(request, response, session
     return;
   }
 
+  if (!session.encryption) {
+    if (!request.headers || (request.headers && request.headers["authorization"] !== this.accessoryInfo.pincode)) {
+      response.writeHead(401, {"Content-Type": "application/hap+json"});
+      response.end(JSON.stringify({status:HAPServer.Status.INSUFFICIENT_PRIVILEGES}));
+
+      return;
+    }
+  }
+
   if (request.method == "GET") {
 
     // Extract the query params from the URL which looks like: /characteristics?id=1.9,2.14,...
@@ -949,14 +967,6 @@ HAPServer.prototype._handleCharacteristics = function(request, response, session
     }.bind(this)), false, session.sessionID);
   }
   else if (request.method == "PUT") {
-    if (!session.encryption) {
-      if (!request.headers || (request.headers && request.headers["authorization"] !== this.accessoryInfo.pincode)) {
-        response.writeHead(401, {"Content-Type": "application/hap+json"});
-        response.end(JSON.stringify({status:HAPServer.Status.INSUFFICIENT_PRIVILEGES}));
-
-        return;
-      }
-    }
 
     if(requestData.length == 0) {
       response.writeHead(400, {"Content-Type": "application/hap+json"});


KhaosT:
For read-only operations, it’s intended to not require the PIN since reading state of accessories shouldn’t be consider as dangerous when it’s in your home network…
Is there a particular reason why you think this change is necessary?

NorthernMan54:
Mostly as my Alexa front end was discovering accessories on homebridge instances that it couldn’t control ( Which isn’t a showstopper).  But also it was just bothering me as a potiental future unnecessary headache if it ever did get dangerous.  Requiring the PIN for all insecure mode operations shouldn’t stop any ‘valid’ user from using the interface.

codyc1515:
Doesn't this defeat the point of the insecure mode? That it's insecure?

oznu:
@codyc1515 - it's still insecure in the sense that you can bypass the HAP encryption.

Insecure mode today still requires the PIN in the header to change the state of accessories, this change would just make GET and PUT requests work the same way.

codyc1515:
I don't really see any reason why not to include it then.

NorthernMan54:
@codyc1515 @oznu for me this is a desired feature as my homebridge-Alexa and node-red modules use auto discovery of hap-node instance.  And currently it finds ones that it can’t control, as it doesn’t know the pin.

Why is this a desired enhancement, I run production and development on the same network and I constantly get accessories I’m developing appear in my production Alexa and node-red

Supereg:
@oznu is this really needed? (a bit late but anyways)
I think this is quite a breaking change. I would rather fully replace this with some more secure and encrypted way of bypassing the HomeKit pairing mechanism (as I already explained to the homebridge org team) and do not really see the point to break this two times in a row.

NorthernMan54:
My underlying rational for putting for the issue was that device discovery found the accessories even if you didn’t have the pin.  And having the pin would limit the response to /accessories to trusted clients.

As we have limited use cases without the pin, fixing is recommended as it exposes the devices.

codyc1515:
Can this not just be brute forced?

Supereg:
> My underlying rational for putting for the issue was that device discovery found the accessories even if you didn’t have the pin.

Wouldn't it be possible from the plugin manufacturers to check if this accessory is controllable (or what is their view)?

> And having the pin would limit the response to /accessories to trusted clients.


> Can this not just be brute forced?

It's called insecure mode cause it's not secure (lol). The pin could also just be sniffed. The packets are transmitted unencrypted. There is really no point of sending the pin at all honestly.

> As we have limited use cases without the pin, fixing is recommended as it exposes the devices.

I would hate to break things for this minor change, which doesn't event better anything in regards to "security". But maybe we could give some timeline, one or two month so plugins can update beforehand, latest versions of plugins can spread and then maybe merge it?

Supereg:
> Can this not just be brute forced?

Any **unpaired** hap-nodejs can also be brute forced easily (relatively). HAP spec states that after 100 unsuccessful attempts you must deny any further attempts to pair. Didn't bother to implement this yet, as we don't really have a nice way of a factory reset, we don't have a hardware button 😅

codyc1515:
Yikes. If it's transmitted in plain-text are we not actually just giving a false impression of security? I don't think that we should do it in that case. Regarding the brute forcing I was more referring to the PIN number that would be sent (is that the same as the HomeKit code)?

Supereg:
> Yikes. If it's transmitted in plain-text are we not actually just giving a false impression of security? I don't think that we should do it in that case. Regarding the brute forcing I was more referring to the PIN number that would be sent (is that the same as the HomeKit code)?

It is the HomeKit code yes.
I aim to replace this whole insecure mode thing with something which does allow other applications access the instance (if enabled), uses pretty much HAP defined protocols (basically the transient pairing mechanism, which isn't currently supported) and thus is as secure as a regular HomeKit made transaction/connection. But this will probably still take some time.

codyc1515:
It just seems like a bad idea if we are going to transmit the HomeKit PIN over plain-text.

> Any **unpaired** hap-nodejs can also be brute forced easily (relatively). HAP spec states that after 100 unsuccessful attempts you must deny any further attempts to pair. Didn't bother to implement this yet, as we don't really have a nice way of a factory reset, we don't have a hardware button 😅

Could we not just increment a count of the number of pairing attempts in memory and once it exceeded 100 reject all pairing attempts? Because it is held in memory it would clear on restart.

Supereg:
> It just seems like a bad idea if we are going to transmit the HomeKit PIN over plain-text.

It's somewhat semi okay, as this "just" happens in the local network, but if you see for what this is used (mostly for plugins like Alexa or google home integration), this developers could well adopt some library to securely call a HAP-NodeJS instance.

> Could we not just increment a count of the number of pairing attempts in memory and once it exceeded 100 reject all pairing attempts? Because it is held in memory it would clear on restart.

Would be a start. HAP has a special error code for this, would be interesting if the Home App shows anything special if it encounters such a bricked accessory.

NorthernMan54:
@codyc1515 @Supereg My rational for this pull request was not to increase the overall security in insecure mode ( which by definition is an oxymoron).  But to give more flexibility to people using insecure mode, and block clients that don’t have the pin.

I found the issue while running multiple “homes” on my local network and found that my Alexa plugin found all instances on the local network even when it did not know the pin.  My secondary home used a different pin as an attempt to separate the two “homes”.  This was more of an isolation method than a security feature.

As the next version has multiple breaking issues, including this should not cause significant issue.

Supereg:
I will close this PR for now.
There isn't currently an intention to add the requested feature.
There is a plan to completely replace the current insecure mode with something more secure (maybe). The idea is, when we do a breaking change we might as well just get the devs ready to implement something "better".

PR #655, opened 2019-03-18
During unpair make sure _advertiser is not null

Shaquu:
In unpair handler I added extra check for _advertiser before we call update on it.
It fixes NRCHKB/node-red-contrib-homekit-bridged#22

In node-red-contrib-homekit-bridged we sometimes have situation like this:

- create bridge with accessory
- publish bridge
- add bridge in Home.app
- remove accessory from bridge and reload bridge
- bridge is not published due to no accessories
- remove bridge from Home.app cause fatal error (publish was not called so we have unassigned _advertiser)


KhaosT:
Thanks 🙏

PR #658, opened 2019-04-05
Check number of bridged accessories

ebaauw:
Throw an error when trying to add more than 99 bridged accessories to a bridge.

It's mostly relevant for homebridge, where a plugin doesn't know how many accessories have already been exposed by other plugins.  This seems the cleanest place to implement the check, though.

### DIFF
--- a/lib/Accessory.js
+++ b/lib/Accessory.js
@@ -16,6 +16,8 @@ var IdentifierCache = require('./model/IdentifierCache').IdentifierCache;
 var bufferShim = require('buffer-shims');
 // var RelayServer = require("./util/relayserver").RelayServer;
 
+const MAX_ACCESSORIES = 99; // Maximum number of bridged accessories per bridge.
+
 module.exports = {
   Accessory: Accessory
 };
@@ -264,6 +266,11 @@ Accessory.prototype.addBridgedAccessory = function(accessory, deferUpdate) {
     var existing = this.bridgedAccessories[index];
     if (existing.UUID === accessory.UUID)
       throw new Error("Cannot add a bridged Accessory with the same UUID as another bridged Accessory: " + existing.UUID);
+  }
+
+  // A bridge too far...
+  if (this.bridgedAccessories.length >= MAX_ACCESSORIES) {
+    throw new Error("Cannot Bridge more than 99 Accessories", MAX_ACCESSORIES);
   }
 
   // listen for changes in ANY characteristics of ANY services on this Accessory


Shaquu:
Why it is 99 and not 100?
Current documentation of HAP states:

> A bridge must not expose more than 100 HAP accesory objects.

ebaauw:
I think the bridge itself counts as one.

KhaosT:
Nice :)

HomeDevices:
Strange, because i have 125 accessories work.

ebaauw:
> Strange, because i have 125 accessories work.

How do you count them?  Note that Apple’s Home app lies about the number of accessories, counting services (tiles) instead.

HomeDevices:
Image

HomeDevices:
This is independedt acessories. Only relays and switches.

HomeDevices:
So if you shut down homekit, it easy to count ;)

ebaauw:
The Home screen counts services (tiles), not accessories.  Also, the limit is the number of accessories behind a single bridge (homebridge instance) - not the total number of independent accessories (accessories that were paired independently to HomeKit).  To check the number of bridged accessories, see the Settings of the Bridge:
IMG_0156

If you add too many, HomeKit becomes unresponsive.

HomeDevices:
Image

HomeDevices:
It works.

HomeDevices:
Maybe old versions only work? And after some version they added limit?

ebaauw:
It looks like the accessories are not responding?  That is what would happen when you exceed the limit.

HomeDevices:
This work about 2 years. I stopped it manually today for some improvements.

HomeDevices:
Image
Running now

KhaosT:
Okay turned out that HomeKit on newer version have this limit to be 150, instead of 100. I'll update it to 149 then.

HomeDevices:
Very interesting if you are right, but it work 2 years with this configuration. So i think this is another problem. I will go to read official homekit datasheet from Apple.

HomeDevices:
Yes, all time it was 150.

ebaauw:
>  I will go to read official homekit datasheet from Apple.

From the HomeKit Accessory Protocol Specification (Non-Commercial Version), R1, section 2.5.3.2:
> A bridge must not expose more than 100 HAP accessory objects.

> Okay turned out that HomeKit on newer version have this limit to be 150, instead of 100.

Cool!  @KhaosT, did you find this in any other documentation, or by analysing the HomeKit runtime code?  Slightly off-topic: do you know if they also increased the CKRecord limit to sync the HomeKit configuration over iCloud?  And number of services per accessory, and number of characteristics per service?

>   I'll update it to 149 then.

Cool!  I messed up the error message, sorry for that.  Will issue a PR with a fix (making sure I've had my coffee first).

To confirm: I setup a homebridge instance with 143 accessories and paired it to my test Home (using a test Apple ID).  Indeed, this seems to work, although the Home app occasionally seems to become unresponsive for half a second or so.  Could very well be caused by the iOS 12.4 public beta 1 or the old iPhone 5s that I use for testing.  I seems to work fine on my iOS 12.2 devices (I share my test Home with my real Apple ID).

KhaosT:
@ebaauw the value was found inside HomeKitDaemon.

HomeDevices:
I think not need add any limit in code, because if it was 100, then up to 150 and later can be other. So not need add limit. Only note about max devices. Each user will add on his own risk.

jvmahon:
FYI - Apple's HAP Specification Version 2.0 (Non-Commercial Version), now states that the limit is 150.  See section 2.5.3.2.  Version 1.0 had stated a limit of 100.

PR #669, opened 2019-05-23
Better support for TypeScript developing

mzyy94:
I defined some missing types to develop using typescript personally.
It's better that these are merged with origin repository here.
Please check that types are correct.

## Changes

- Add types of PredefinedService using following command
`egrep -ho "Service\\..+\\.UUID" lib/gen/HomeKitTypes* | cut -d '.' -f 2 | sort | uniq | xargs -n1 -I{} echo "{}: PredefinedService;" `
-  Add types of Characteristic using following command
`egrep -ho "Characteristic\\..+\\.UUID" lib/gen/HomeKitTypes* | cut -d '.' -f 2 | sort | uniq | xargs -n1 -I{} echo "{}: Characteristic;"`
- Set types of `interface HAPNodeJS` properties
- Add type-safe callback function `type`
- Replace `boolean | string | number` to `type CharacteristicValue`


KhaosT:
Nice ❤

PR #673, opened 2019-06-09
Fix and improve typescript definition

mzyy94:
It has wrong type that I previously pull-requested #669. And missing Service prototypes are added.


jonavila:
@mzyy94 Thanks for the improved definitions. Until these get released, I've been using https://github.com/ds300/patch-package to get these in my plugin.

mzyy94:
cool

PR #678, opened 2019-06-24
Special case for AccessoryInformation service

### LINKED ISSUE #677: Service.AccessoryInformation with displayName thows same UUID error
When creating a new `AccessoryInformation` service with a displayName argument, I receive the following error:

```Error: Cannot add a Characteristic with the same UUID as another Characteristic in this Service: 00000023-0000-1000-8000-0026BB765291```

It seems like this issue stems from the fact that the Service constructor will create the *Name* characteristic for every service when the displayName is provided:

https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/Service.js#L58

Subsequently, the AccessoryInformation service will attempt to add the same *Name* characteristic that has already been added:

https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/gen/HomeKitTypes.js#L2660

I'm building a platform plugin using Homebridge & Typescript and I'm wondering what's the correct way to add the AccessoryInformation service to all my accessories. 

I believe the only workaround at the moment is to omit and the displayName when creating the AccessoryInformation service and to make sure the accessory has a `name` property defined.

FYI @KhaosT

jonavila:
Avoid adding the `Name` characteristic to `Service.AccessoryInformation` since it will be added when defining that service. This will prevent the same UUID as another Characteristic in this Service error.

Fixes #677

### DIFF
--- a/lib/Service.js
+++ b/lib/Service.js
@@ -52,6 +52,10 @@ function Service(displayName, UUID, subtype) {
   // if one was given
   // if you don't provide a display name, some HomeKit apps may choose to hide the device.
   if (displayName) {
+    // Don't add the `Name` characteristic to `Service.AccessoryInformation` since it will be added when defining that service.
+    if (UUID === '0000003E-0000-1000-8000-0026BB765291') {
+      return;
+    }
     // create the characteristic if necessary
     var nameCharacteristic =
       this.getCharacteristic(Characteristic.Name) ||


tobekas:
This workaround is problematic, because new Services might be added, which also have a requirement Characteristic.Name. See #677
The other thing is, Name‘s default value would be lost either.

jonavila:
@tobekas Yeah, that's possible about future Services adding a required Characteristic.Name. As far as I can tell, the workaround of automatically adding the name Characteristic.Name to every Service in the constructor was done to prevent certain homekit apps from not displaying the service. I personally think that it shouldn't be added by default, but that would be a much bigger since every project that depends on hap-nodejs would have to account for that change.

On your second point, the AccessoryInformation service is added anyways to every single Accessory as seen here:

https://github.com/KhaosT/HAP-NodeJS/blob/master/lib/Accessory.js

That's set the Name characteristic to the displayName of the accessory. Perhaps, an AccessoryInformation service should never be created manually, since it's added automatically by all accessories? I guess that argument could be made. In the case of the Homebridge project, it overwrites the other characteristics like Manufacturer, Model, SerialNumber... with the ones specified in your own version of AccessoryInformation.

https://github.com/nfarina/homebridge/blob/master/lib/server.js#L454

@nfarina FYI. 

Thoughts?

Supereg:
I don't think it's good practice either to just manually add exceptions into the constructor. Especially when only referencing a plain UUID in the code. My plan currently would be do someday rework the execution order so this won't be a problem anymore.
As of this, will close this PR for now.

PR #661, opened 2019-04-13
Update Fan_accessory.js

abilops:
Fixed spelling

PR #685, opened 2019-07-03
Fix typo

Fonger:
apply #661 fix for AirConditioner_accessory.js

Maybe this example should be updated to use Service.HeaterCooler


KhaosT:
Thanks :)

PR #687, opened 2019-07-05
 catch oldValue in setValue directly at the beginning

Apollon77:
Catch oldValue in setValue directly at the beginning to make sure cha…nge events are triggered as soon as value change.

Reason: I found some Homebridge Plugins that break the whole "sending change events" process by setting the value itself in the set-handler. The result is that old and new value are the the same and no change event is triggered at all.

Example: https://github.com/YinHangCode/homebridge-mi-aqara/blob/master/parser/GatewayParser.js#L267

This change would fix this and really compare oldValue from timepoint of the setValue call with the new value
@KhaosT


KhaosT:
Thanks!

PR #696, opened 2019-08-11
 update definitions for listening to get/set events

kejadlen:
Giving Typescript some help so I don't have to manually add types for the `set` callback. I also had to make `value` optional in the `get` callback for the error case, even though I thought Typescript did that by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.4.53",
+  "version": "0.5.0",
   "description": "HAP-NodeJS is a Node.js implementation of HomeKit Accessory Server.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",


hassankhan:
Hi @kejadlen, is this PR still needed? v0.5 is now rewritten in Typescript, and I believe the callback parameter issue is now resolved 👍

kejadlen:
Probably? I’ll need to check, but likely won’t be able to get to it until the weekend.

kejadlen:
Oops, closed this by accident. Need to update all the types and such in my library to see if this compilation error is still happening.

kejadlen:
Still seeing issues with type inference for get/set:

```
src/accessorizer.ts:24:61 - error TS7006: Parameter 'callback' implicitly has an 'any' type.

24             characteristic.on(CharacteristicEventTypes.GET, callback => {
                                                               ~~~~~~~~

src/accessorizer.ts:33:62 - error TS7006: Parameter 'value' implicitly has an 'any' type.

33             characteristic.on(CharacteristicEventTypes.SET, (value, callback) => {
                                                                ~~~~~

src/accessorizer.ts:33:69 - error TS7006: Parameter 'callback' implicitly has an 'any' type.

33             characteristic.on(CharacteristicEventTypes.SET, (value, callback) => {
                                                                       ~~~~~~~~
```

hassankhan:
Hi @kejadlen, I'm not entirely sure that I can fix those ... as I understand it, the Typescript compiler requires you to specify types for those parameters as well. I believe you can switch those errors off with the `--noImplicitAny` compiler flag.

kejadlen:
I was able to get it working with the older type definitions, so I figure that maybe it's possible in "pure" Typescript too? I haven't dug too much into it, but this is how I got type inference to work better pre-0.50: https://github.com/kejadlen/HAP-NodeJS/commit/32e7fb7111e834d70d5cd8d9e61433ba501388e1

PR #700, opened 2019-08-22
Set pv bonjour record to 1.1

Supereg:
Accessory Tester complains about protocol version in bonjour txt:
```
Error: IP_NonCompliantBonjourTXTRecord: Bonjour TXT record for "device-xxxx" is invalid.
Protocol Version (pv) is incorrect, and it must be set to 1.1.
```


codyc1515:
Did you end up implementing this?

PR #691, opened 2019-07-28
Refactor to typescript

hassankhan:
Hi there,

As part of homebridge#2266, I wanted to give this project a bit of Typescript love too. 

I've converted everything except the `lib/gen/import.js`, since I'm not quite sure if that file will continue to live on. I tried running the script, and it didn't generate as many devices as the existing file had 😕 

I've also added Jest and started writing some very basic unit tests, which I'll add on later.

PR #701, opened 2019-08-24
Fix minor breakages in v0.5

hassankhan:
This is a follow-up PR for any issues/breakages introduced by #691


KhaosT:
Thanks :)

Supereg:
Seems still broken. When running HAP-NodeJS using `ts-node --files src/Core.js` it doesn't actually publish any accessories from the `accessories` directory.
Only output is `HAP-NodeJS starting...` after which the program terminates.
Seems to have some sort of issues in the AccessoryLoader.

PR #702, opened 2019-08-26
Reimplement /pairings route to match the HAP spec

Supereg:
**Reimplemented /pairings route to correctly match the HAP spec:**
- added permission management for paired controllers
- validate add/remove pairing request:
  - check controller admin permission
  - do not allow overriding of public key (add)
- add support for list pairings requests
- unpair every regular user controller when last admin controller unpairs
- destroy session if controller unpairs itself or any currently connected controller gets unpaired

<br>
Also in this PR:

- completed list of tlv values and status codes
- matched error response of pair-verify process to the expected behavior according to the HAP spec

<br>
<br>
Everything in this PR is tested and fulfills the expectations of the HAP spec :)
Happy to answer any questions concerning this PR

### DIFF
--- a/src/lib/model/AccessoryInfo.ts
+++ b/src/lib/model/AccessoryInfo.ts
@@ -4,6 +4,18 @@ import tweetnacl from 'tweetnacl';
 import bufferShim from 'buffer-shims';
 
 import { Categories } from '../Accessory';
+import { Session } from "../util/eventedhttp";
+
+export enum PermissionTypes {
+  USER = 0x00,
+  ADMIN = 0x01, // admins are the only ones who can add/remove/list pairings (also some characteristics are restricted)
+}
+
+export type PairingInformation = {
+  username: string,
+  publicKey: Buffer,
+  permission: PermissionTypes,
+}
 
 /**
  * AccessoryInfo is a model class containing a subset of Accessory data relevant to the internal HAP server,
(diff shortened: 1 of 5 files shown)


KhaosT:
Thanks for PR 🙏 I'll probably have some time to review this next Saturday.

Supereg:
One thing I'm not quite sure about is if the admin permission of controllers is used for any other routes besides the `/pairings` route. But when looking at the spec (the non-commericial one is pretty outdated though) it seems this is currently the only use case.

KhaosT:
@Supereg I think there are a few characteristics restricted operation to only be performed by admin sessions, but I guess it doesn't really matter here 😅

Supereg:
Ah okay. Maybe somebody can figure that out at some point. We were fine until now with not saving the permission at all, so that is probably not highest priority 😅😂

netRunner0:
How to check it? Need to replace js files, or put ts next?

src/lib/model/AccessoryInfo.ts
```diff
@@ -4,6 +4,13 @@ import tweetnacl from 'tweetnacl';
 import bufferShim from 'buffer-shims';
 
 import { Categories } from '../Accessory';
+import { Session } from "../util/eventedhttp";
+
+export type PairingInformation = {
+  username: string,
+  publicKey: Buffer,
+  permission: number, // 0x00: regular user, 0x01: Admin (able to add/remove/list pairings)
```
hassankhan: Could we make this an `enum` please?

Supereg: ✅

Supereg:
Any updates on this?

KhaosT:
Nice! Merged :)

PR #703, opened 2019-08-26
Fixed: AccessoryLoader now loads typescript files correctly

Supereg:
As mentioned in #701 typescript files would be ignored by the AccessoryLoader after the refactoring to typescript. Now AccessoryLoader will try to load javascript as well as typescript files.


KhaosT:
Thanks :)

PR #704, opened 2019-08-26
Fixed: resource/snapshot requests for bridged cameras now work as intended

Supereg:
When HomeKit invokes a `/resource` request for a camera, which is not the primary accessory (for example for bridged cameras and maybe also multiple cameras in one bridge), HomeKit includes the `aid` property in the request to specify from which accessory the snapshot should be gathered. This property is currently ignored by HAP-NodeJS and as a result any sort of snapshot requests won't work for bridged cameras.
This PR fixes that issue.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -109,6 +109,7 @@ export enum ResourceTypes {
 }
 
 export type Resource = {
+  'aid'?: number;
   'image-height': number;
   'image-width': number;
   'resource-type': ResourceTypes;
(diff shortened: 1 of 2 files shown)


Supereg:
Don't know if there are any open issue regarding this issue. Didn't find any after a quick search.

KhaosT:
Sweet :) The initial implementation on HomeKit side doesn't supply aid so it wasn't built in here. Glad to know they updated the implementation to support bridged cameras.

Supereg:
> Sweet :) The initial implementation on HomeKit side doesn't supply aid so it wasn't built in here. Glad to know they updated the implementation to support bridged cameras.

Stumbled across this while debugging my rewritten personal camera setup on my pi. They probably needed to support things like the Arlo cameras. Didn't know this wasn't supported from the start on.

fennec622:
Hi 
Can you confirm that resolv snapshot on motion detect ?

https://github.com/KhaosT/homebridge-camera-ffmpeg/issues/351

If yes how can i do for update hap-nodeJS in homebridge

thanks for your job

Supereg:
@fennec622 this fixes only a certain scenario. Snapshot wouldn’t work for cameras published by an HomeKit bridge. The snapshot wouldn’t work at all. So I doubt that this has something to do with your issue with motion sensors.

fennec622:
Ok thanks

PR #706, opened 2019-08-31
 Implement /characteristics route according to the spec

Supereg:
* Implement http response according to the spec:
  * for writes:
    * 204 "No Content" if no errors occurred
    * 207 "Multi-Status" if at least one characteristic reports an error or returns a write response
  * for reads:
    * 200 if no errors occurred (also do not include the `status` property if `value` was returned successfully)
    * 207 "Multi-Status" if at least one characteristic reports an error
* Added check to ensure operations performed on characteristics are actually permitted

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -8,7 +8,8 @@ import { Service, ServiceConfigurationChange, ServiceEventTypes } from './Servic
 import {
   Characteristic,
   CharacteristicEventTypes,
-  CharacteristicSetCallback
+  CharacteristicSetCallback,
+  Perms
 } from './Characteristic';
 import { Advertiser } from './Advertiser';
 import { HAPServer, HAPServerEventTypes, Status } from './HAPServer';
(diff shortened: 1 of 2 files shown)


src/lib/Accessory.ts
```diff
@@ -822,6 +823,20 @@ export class Accessory extends EventEmitter<Events> {
...
+      if (!characteristic.props.perms.includes(Perms.PAIRED_READ)) { // check if we are allowed to read from this characteristic
+        debug('[%s] Tried reading from Characteristic which does not allow reading (iid of %s and aid of %s)', this.displayName, characteristicData.aid, characteristicData.iid);
+        const response: any = {
+          aid: aid,
+          iid: iid
+        };
+        response[statusKey] = Status.WRITE_ONLY_CHARACTERISTIC;
+        characteristics.push(response);
+
+        if (characteristics.length === data.length)
```
KhaosT: Can you wrap single line if statement in curly brackets?

Supereg: Yes. Saw the same code style above so wasn't sure what style is preferred.

netRunner0:
How to check it? Need to replace js files, or put ts next?

KhaosT:
Sweet, thanks for the PR 🙏

PR #710, opened 2019-09-13
Publishing accessory type icons for pairing

n0rt0nthec4t:
Small change to allow overriding of the default type icon type which is displayed in the HomeKit pairing window. By default, accessories are published with the category type as "OTHER". This change will allow uses to override this type in their accessory code


KhaosT:
Thanks for the PR 🙏

PR #712, opened 2019-09-18
Code quality and contributor docs

hassankhan:
Just wanted to follow up with some code standards and documentation.

Would love feedback on the contributor documentation, especially the migration path.

/cc @n0rt0nthec4t @KhaosT @Supereg


n0rt0nthec4t:
I think a detailed guide is needed for installation now given the vast increase in dependency’s needed for ts-node.. I’m still struggling with that side

Supereg:
Looked at the README and the CONTRIBUTING markdown files. Seems okay. One thing I don't quite understand why you moved the "API" section into the CONTRIBUTING and removed it completely from the README. I don't think people who just want to write their own accessories gonna look into the CONTRIBUTING file.
In general I think explanation and assistance could be improved for new developers.

CONTRIBUTING.md
```diff
@@ -0,0 +1,129 @@
...
+exposed as a separate subclass in [HomeKitTypes](src/lib/gen/HomeKit.ts).
+
+See each of the corresponding classes for more explanation and notes.
+
+## Notes
+
+Special thanks to [Alex Skalozub][link-alex-skalozub], who reverse-engineered
+the server side HAP. ~~You can find his research [here][link-homekit-research]
+.~~ (Sadly, on Nov 4, Apple sent the [DMCA][link-apple-dmca] request to Github
+to remove the research.)
```
KhaosT: Maybe put this section back to README?

hassankhan:
Hi guys, sorry for the delay, been super busy with work projects lately. I've tried my best to clear some things up. If there were a list of specific issues users may have, I think it'd be great to add them to the `MIGRATION.md` file.

I have enabled edits from maintainers on the PR so at the very least @KhaosT should be able to fix up the PR if I'm unavailable.

ACCESSORIES.md
```diff
@@ -0,0 +1,41 @@
...
+## HomeKit Protocol
+
+Hint: the Homekit Application Protocol (HAP) allows that you can pair a Homekit
+device with one device. As soon as the Homekit device is paired, its not
+possible to pair with another iOS device anymore.
+
+## API
+
+HAP-NodeJS provides a set of classes you can use to construct Accessories
+programatically. For an example implementation, see [Lock_accessory.ts](src/accessories/Lock_accessory.ts).
```
Supereg: typo: programmatically

Supereg:
We have one little problem with this current configuration of husky and eslint.
We have a huge code base wich is not compliant with the supplied eslint settings 😅 

When a user just edits a little thing in a file like `BridgeCore.ts` (and even assuming those changes comply with the requirements of eslint) and `lint-staged` gets executed the whole file is getting altered by eslint/prettier (Replacing double quotes with single quotes, inserting trailing `,` it json, inserting missing `;`, removing or adding some whitespaces). The problem here isn't really that eslint/prettier is executed with the --fix/--write parameter, rather, that in the current state of the project, it will also fix everything else of the file, even stuff that the user hasn't touched. Which then results in pull requests which are full of this. 

Furthermore users would find themselves unable to commit anything at all, since there are so many eslint errors in the current code base which cannot be fixed automatically by eslint. So altering a little comment in a file like `HAPServer.ts` will require fixing over 100 eslint errors.

And lastly I would like to discuss some options, as they are not really applicable for this project or aren't really that good according to my opinion:
- `@typescript-eslint/explicit-function-return-type` - Missing return type on function (warning): Many of our functions are callback driven. This option will fall every of them 
- `@typescript-eslint/explicit-member-accessibility` - Missing accessibility modifier on class property Types (error): Do we really want to specify `public` access modifier for every property? After all this isn't Java lol
- `@typescript-eslint/no-non-null-assertion` - Forbidden non-null assertion (error): Forbids the `!` postfix operator. This is a bit tricky, as in some cases a warning could be good, but in some cases the operator makes sense.
- `@typescript-eslint/prefer-interface` - Use an interface instead of a type literal (error): Is this really necessary to replace all type defs with interfaces?


To sum up. I think the linting changes are good and with the project growing they are definitely needed. However the current project should be immediately altered to comply with those eslint/prettier restrictions (don't know how to handle that exactly, @KhaosT ?), otherwise this would create only more problems for contributors.
I don't know what to feel about automatically altering and fixing linting errors when committing. I'm personally not a big fan of the idea that I could commit something I don't know about (don't know if the automatic correction of linting errors could potentially create any issues?). But that could be just me.
Regarding the linting/prettier options, I don't know what's the consensus is there (tabWidth, single quotes only, restricting printWidth, etc; currently they are all over the place).

P.S. I think the documentation stuff is pretty much ready, and stuff like the MIGRATION.md would be great to get out fast, maybe create a separate PR for this? The code quality stuff probably needs still some changes (?) and the new documentation could then be merged faster by @KhaosT(?).

ebaauw:
Unfortunately, there's multiple standards out there for Javascript linting, which aren't compatible with each other (e.g. I use `standard`, which removes the training `;`s instead of inserting them).  I haven't found any compelling argument to use one over the other, so I just picked one.

I don't know if the same holds for Typescript, but I would stronly suggest to pick one style and stick to it.  Cherry picking individual options is a nightmare for contributors.  Or, to quote from the   Ten Commandments for C Programmers:
> 8. Thou shalt make thy program's purpose and structure clear to thy fellow men by using the One True Brace Style, even if thou likest it not, for thy creativity is better used in solving problems than in creating beautiful new impediments to understanding.

Supereg:
A CONTRIBUTOR.md was added to the latest [beta](https://github.com/homebridge/HAP-NodeJS/tree/beta). It was heavily inspired by the one provided in this PR.

HAP-NodeJS will move ESLint in the near future and will use basically the same style as it was introduced in the homebridge project with the latest typescript rewrite.

Lint checking will be done using GitHub Actions; we won't install any git hooks.

As of this I will close this PR. Thanks for your contribution. Even if it wasn't exactly accepted, it help implementing proposed changes very much and many parts of it ended in the current code base.

hassankhan:
@Supereg No worries, thanks for getting it over the finish line, massively appreciate all your work fixing the rough edges from the Typescript conversion a while back 👍

PR #718, opened 2019-10-01
init function - optional parameter storagePath

Supereg:
Checking the content of the init function I assume the parameter should be an optional.

~ Andi

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,7 @@ export * from './lib/util/tlv';
 
 export * from './types';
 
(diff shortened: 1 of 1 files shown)


Supereg:
@hassankhan as you were the person who refactored the whole project. Could you verify that the current version is a mistake? ☺️

KhaosT:
We have the case to handle when the input is undefined so I think it make sense ^^

hassankhan:
Sorry for the late reply, that indeed was an oversight on my part 😅

PR #719, opened 2019-10-08
Fixing session dictionary getting messed up handling multiple controllers of the same Apple ID

Supereg:
Introduced with my PR in #702 the `sessions` variable in the Session class tracked all connected clients by the `username`. The `username` identifies every user connected to the home uniquely. As a result the `sessions` dictionary will get messed up with multiple Apple devices (from the same user/Apple ID) connecting and disconnecting from the HAP server.

Even though I knew that this id identifies every Apple ID connected to the home I messed that one up. Sorry for that. Here is the fix 😅

**Impact:**
To anyone who runs the current version and to clarify the impact. When you remove anybody from your home, the current HAP-NodeJS will have problems to kill any currently open connections from the user you are removing. Which means if devices from a persons in your home are currently connected to your HAP-NodeJS driven accessory, they may be able to still control your devices just after you remove the from your Home until their devices close the connection.
To workaround this issue just reboot your HAP-NodeJS instance after you removed the person.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -27,6 +27,7 @@ import {
 } from '../types';
 import { Camera } from './Camera';
 import { EventEmitter } from './EventEmitter';
+import { Session } from "./util/eventedhttp";
 
 // var HomeKitTypes = require('./gen/HomeKitTypes');
 // var RelayServer = require("./util/relayserver").RelayServer;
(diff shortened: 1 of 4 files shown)


KhaosT:
Thanks for catching the issue... I should review it more careful the first time 😅

Supereg:
> Thanks for catching the issue... I should review it more careful the first time 😅

Probably 😄  Don't trust me 😂😅

Fortunately the impact on session security isn't that huge.

PR #722, opened 2019-10-18
Update Advertiser.ts

n0rt0nthec4t:
Small change to create more "commercial" like accessory advertising names in the HomeKit pairing window. Replaces '-' with ' '

before:
SomeAccessory-X1Y2

after:
SomeAccessory X1Y2

### DIFF
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -56,7 +56,7 @@ export class Advertiser {
      */
     var host = this.accessoryInfo.username.replace(/\:/ig, "_") + '.local';
     var advertiseName = this.accessoryInfo.displayName
-      + "-"
+      + " "
       + crypto.createHash('sha512').update(this.accessoryInfo.username, 'utf8').digest('hex').slice(0, 4).toUpperCase();
 
     // create/recreate our advertisement


Supereg:
Just to throw it in there as a note.

HomeKit expects the name published through Bonjour to be the same as the one set in the Accessory Information Service. This is currently not the case since the hash is appended to the Bonjour name entry to ensure uniqueness on the network.

I don't know if this is known, I don't know if this affects anything (on the HomeKit controller side), I don't know how ambitious this project is to be 100% on par with the spec and I also don't know what the opinions would be about this change.

@KhaosT

n0rt0nthec4t:
> Just to throw it in there as a note.
> 
> HomeKit expects the name published through Bonjour to be the same as the one set in the Accessory Information Service. This is currently not the case since the hash is appended to the Bonjour name entry to ensure uniqueness on the network.
> 

I had thought of proposing removing the hash after the accessory name to align to the way the HomeKit simulator advertises names

KhaosT:
Yeah I don't feel strongly either way... Since we can't put MFi authentication coprocessor in here, it won't be spec compliant anyway 😅

n0rt0nthec4t:
If thats the case, should I do another pull request with a change to remove the has value and go for the accessory name only as per what the HomeKit accessory simulator does?

PR #723, opened 2019-10-20
Update to protocol version 1.1.0

Supereg:
This PR makes the following changes:

- The ProtocolInformation service gets added to the primary accessory
- The `/prepare` route got implemented to support prepared writes

<br>
Note:

I'm unsure what exactly the benefits are for prepared writes as no characteristics require it at this point. However I discovered that for lock services a HAP controller executes a prepared write when opening/closing. I don't know to which services this also applies.


KhaosT:
Yeah I think right now it's primarily for lock related services to prevent attacker from holding TCP write request to security related characteristics and send those later.

Issue #33, opened 2015-01-13
Build Issue with Curve


brianantonelli:
On a fresh Raspbian install (following the Raspi install doc) I'm running into an error when building Curve. I am able to successfully install node-gyp, but when I run npm rebuild it errors out on the curve module. It's worth noting that if I do a node-gyp rebuild from the curve module it builds fine. 

root@raspberrypi:/home/pi/HAP-NodeJS# npm rebuild

> curve25519@1.0.0 install /home/pi/HAP-NodeJS/node_modules/curve25519
> node-gyp rebuild

gyp ERR! clean error 
gyp ERR! stack Error: EACCES, unlink 'build'
gyp ERR! System Linux 3.12.35+
gyp ERR! command "node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /home/pi/HAP-NodeJS/node_modules/curve25519
gyp ERR! node -v v0.10.35
gyp ERR! node-gyp -v v1.0.1
gyp ERR! not ok 

npm ERR! curve25519@1.0.0 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the curve25519@1.0.0 install script.
npm ERR! This is most likely a problem with the curve25519 package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls curve25519
npm ERR! There is likely additional logging output above.
npm ERR! System Linux 3.12.35+
npm ERR! command "/usr/local/bin/node" "/usr/local/bin/npm" "rebuild"
npm ERR! cwd /home/pi/HAP-NodeJS
npm ERR! node -v v0.10.35
npm ERR! npm -v 1.4.28
npm ERR! code ELIFECYCLE
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /home/pi/HAP-NodeJS/npm-debug.log
npm ERR! not ok code 0

PR #725, opened 2019-10-22
HomeKit accessory types

n0rt0nthec4t:
Discovered new HomeKit accessory type, #33, which appears as a router image in the HomeKit pairing window. This change defines type 33 as "ROUTER"

IMG_1322


codyc1515:
Unfortunately there is no documentation as yet on the Characteristics for this service.

apexad:
> Unfortunately there is no documentation as yet on the Characteristics for this service.

Given that though, is there any reason not to at least add it to the list like this PR does? 

Otherwise if the goal is to wait until Characteristics are known and a more complete PR is created, this one should be closed.

Issue #721, opened 2019-10-10
Temperature in decimal format


kajaba:
Hello 
I have a problem with decimal format in temperature characteristic. I’m trying to set, for example, value 38.5 and all I can get in the HomeKit app is 39

> sensor
> .addService(Service.TemperatureSensor)!
> .getCharacteristic(Characteristic.CurrentTemperature)!
> .on(CharacteristicEventTypes.GET, (callback: NodeCallback<CharacteristicValue>) => {
>     // return our current value
>     callback(null, 38.5);
>   });

Am I doing something wrong?

Thank you very much for helping

codyc1515:
Appears to be an iOS 13 thing, not sure if it’s a bug or not. Have you tried asking Siri to set it with a decimal?

kajaba:
I think that I had these issues also in previous versions of iOS. 
I could ask Siri what is the temperature, but in the code I am setting the temperature manually and somewhere between node Hap server and iOS homekit app (maybe the app has this bug) is this value rounded to whole number (int)

jvmahon:
Just a guess, but try adding
    .setProps({ minStep:0.1})

To your TargetTemperature and CurrentTemperature characteristics.

kajaba:
I think that CurrentTemperature characterist is well configured:
`export class CurrentTemperature extends Characteristic {

  static readonly UUID: string = '00000011-0000-1000-8000-0026BB765291';

  constructor() {
    super('Current Temperature', CurrentTemperature.UUID);
    this.setProps({
      format: Formats.FLOAT,
      unit: Units.CELSIUS,
      maxValue: 100,
      minValue: 0,
      minStep: 0.1,
      perms: [Perms.READ, Perms.NOTIFY]
    });
    this.value = this.getDefaultValue();
  }
}
`

SystemKeeper:
I stumbled across the same issue and I think its because of an explicit integer conversion at https://github.com/KhaosT/HAP-NodeJS/blob/e87f7743c8f42d5f210a6fabf15591567c665b64/src/lib/Characteristic.ts#L528
Looking at older revisions (before moving to TS) it looked like this:
https://github.com/KhaosT/HAP-NodeJS/blob/c11776dc4a145381228ecfadb47dd583977e367d/lib/Characteristic.js#L285

Any idea why this was added? After removing the parseInt() I can set temperatur to 22.5 (22.3 will be displayed as 22.5, but guess thats not a problem of this repo!?)

[...]

PR #729, opened 2019-10-27
Fixed: floating point values are now validated correctly again

Supereg:
This PR fixes a regression introduced in bb0c5b9af02e.
Issue discussed in #721.


KhaosT:
Oops

PR #731, opened 2019-10-28
Additional characteristics response properties, additional authorisation data and custom handling of non-HAP requests

samuelthomas2774:
#### Additional characteristics response properties

`GET /characteristics` now responds with full characteristic details if the client requests that. I don't know when iOS ever does this, but it's required by the HAP specification.

#### Pass the characteristic write data to characteristic set handlers

This allows characteristics to use additional authorisation data.

```ts
function checkAuthData(authData: string | undefined, value: any) {
    // Check the authorisation data is valid and the user should be able to set this value...
    authData = authData && Buffer.from(authData, 'base64');
    return true;
}

characteristic.on(CharacteristicEventTypes.SET, (value: any, callback: CharacteristicSetCallback, session?: Session, writeData?: CharacteristicWriteData) => {
    // Check additional authorisation data
    // If writeData isn't set then setValue wasn't called by a HAP client - here we assume that the caller is privileged and can make whatever changes they like
    if (writeData && !checkAuthData(writeData.authData, value)) return callback(new Error(Status.INSUFFICIENT_AUTHORIZATION));

    // Continue setting the characteristic...
});
```

#### Pass the write response back to the accessory instead of setting the value

The current behaviour can affect characteristics that can be read normally (the PAIRED_READ permission). Characteristic set handlers can set the current value if that's the desired behaviour.

#### Custom handling of non-HAP requests

This could be used to add a custom page explaining what the server is, instructions for pairing with the server, or a settings page.

```ts
const app = express();

accessory._server!.on(HAPServerEventTypes.UNHANDLED_REQUEST, (req: http.IncomingMessage, res: http.ServerResponse, session: Session) => {
    // Handle this request
    // Since req and res are just normal Node.js HTTP request/response objects we could forward this to an Express application
    app.handle(req, res);
});
```

#### Change HAPServer events and characteristic get/setValue functions to use only the Session object instead of some of it's properties

This simplifies some event handler signatures. Currently we have some events like `(data: any, events: CharacteristicEvents, callback: NodeCallback, session: Session)` where `events === session._connection._events`. This changes events like that to something like `(data: any, session: Session, callback: NodeCallback)`.

This shouldn't break anything that depends on HAP-NodeJS unless they're using HAPServer directly or using the `connectionID` argument of characteristic get/set handlers.


src/lib/Characteristic.ts
```diff
@@ -88,8 +90,8 @@ export type CharacteristicSetCallback = (error?: Error | null, value?: Character
 
 type Events = {
   [CharacteristicEventTypes.CHANGE]: (change: CharacteristicChange) => void;
-  [CharacteristicEventTypes.GET]: (cb: CharacteristicGetCallback, context?: any, connectionID?: string) => void;
-  [CharacteristicEventTypes.SET]: (value: CharacteristicValue, cb: CharacteristicSetCallback, context?: any, connectionID?: string) => void;
+  [CharacteristicEventTypes.GET]: (cb: CharacteristicGetCallback, session?: Session) => void;
+  [CharacteristicEventTypes.SET]: (value: CharacteristicValue, cb: CharacteristicSetCallback, session?: Session, writeData?: CharacteristicWriteData) => void;
```
Supereg: You are breaking plugins with that change.

samuelthomas2774: I'm not sure about the Characteristic changes. Removing `connectionID?: string` likely will break some plugins. I don't know if changing `context: any` to `session?: Session` could break anything though. If that doesn't break anything then could it be changed to this?

```ts
[CharacteristicEventTypes.GET]: (cb: CharacteristicGetCallback, session?: Session, connectionID?: string) => void;
[CharacteristicEventTypes.SET]: (value: CharacteristicValue, cb: CharacteristicSetCallback, session?: Session, connectionID?: string, writeData?: CharacteristicWriteData) => void;
```

src/lib/Characteristic.ts
```diff
@@ -88,8 +90,8 @@ export type CharacteristicSetCallback = (error?: Error | null, value?: Character
 
 type Events = {
   [CharacteristicEventTypes.CHANGE]: (change: CharacteristicChange) => void;
-  [CharacteristicEventTypes.GET]: (cb: CharacteristicGetCallback, context?: any, connectionID?: string) => void;
-  [CharacteristicEventTypes.SET]: (value: CharacteristicValue, cb: CharacteristicSetCallback, context?: any, connectionID?: string) => void;
+  [CharacteristicEventTypes.GET]: (cb: CharacteristicGetCallback, session?: Session, connectionID?: string) => void;
+  [CharacteristicEventTypes.SET]: (value: CharacteristicValue, cb: CharacteristicSetCallback, session?: Session, connectionID?: string, writeData?: CharacteristicWriteData) => void;
```
samuelthomas2774: I'm not sure if this breaks anything so it still needs reviewing. Homebridge plugins aren't aware of their accessory IDs and `keepalive` is always true so there's nothing plugins can read from the events object. The only way I think this could break anything is if a plugin overwrites any Session properties (sessionID, events, encryption, srpServer, username, timedWritePid, timedWriteTimeout, establishSession, _connectionDestroyed).

KhaosT:
Apologies for the delay, the work life got really crazy recently and since this PR changed characteristic write logic quite a bit, I'll need to find a time to go over it more carefully 😅

Supereg:
The latest beta release `v0.9.0-beta.13` includes pretty much all changes proposed here.

HAP-NodeJS will now properly reply perms, type, metaData or ev state on read request if requested via query parameters.

Support for handling additional authorization data was added, though with a little different api https://github.com/homebridge/HAP-NodeJS/blob/e16a3e450d3353624879e5d3b6db225e8a2c260c/src/lib/Characteristic.ts#L477-L498

> Pass the write response back to the accessory instead of setting the value

This was also changed.

> Change HAPServer events and characteristic get/setValue functions to use only the Session object instead of some of it's properties

The signature of get and set handlers was changed so the session object is included instead of the session identifier. The `events` aka `context` field was kept to provide greater backwards compatibility.

Support for handling custom routes was deliberately not added.


Although your specific pull request was not added, huge thanks for contribution as it pointed out those major pain points.

PR #732, opened 2019-10-28
Enforce usage of valid deviceId/username

Supereg:
Added assert to check if supplied username matches the mac address pattern.
Additionally added test to verify that behavior.


KhaosT:
🙏

PR #735, opened 2019-11-03
Pairing/authentication fixes

samuelthomas2774:
- Fix an issue where clients are able to make authenticated requests without completing the pair verify procedure
- Fix an issue where a controller paired to multiple servers running in the same process and is unpaired from one of them is disconnected from all servers
- Track pair setup/verify state so a client can't send multiple pair setup/verify requests and out of order requests


KhaosT:
Nice ^^ Thanks for fixing this 🙏

PR #736, opened 2019-11-03
Implemented the HomeKit Data Stream protocol

Supereg:
This is part one of two PRs (Spoiler alter: part two will add support for Siri remotes. I'm finishing the last bits on that one).

This PR implements full support for HomeKit Data Streams.
The API is currently intended as follows:
* Step 1: You would create a new instance of the `DataStreamManagement` class and add the service returned by `getService()` to your accessory
* Step 2: finished

If an HomeKit Hub detects a combination of supported services it will create a connection.

You could then create some listeners for certain event messages with the respective protocol name (for example the `whoami` event message of the `targetControl` protocol. This is what the SiriRemote uses to identify DataStream sessions) or listen to more general events like `CONNECTION_OPENED` or `CONNECTION_CLOSED` to get a reference to an open data stream connection.

Once you got a reference to an `DataStreamConnection` you can add protocol handlers to that specific connection or send event, request or response messages.

As the HomeKit DataStream protocol supports different sized ints (int8, int16, int32, int64) I introduced some wrapper classes in `DataStreamParser.ts` (and also for some other data types) to mark how to encode the javascript number type. By default a number is encoded in the smallest possible type. It is ugly but I though it is the least ugly solution.
Also, as we have a minimum requirement of node v4 for the project currently, writing int64 isn't really supported. And also reading is a bit hacky but it seems to work (however we only really support 53 bit ints). The javascript bigint type is only support for node v10. The end goal would be to move to that data type (maybe?).
But feedback for any better solution to this problem is welcomed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ export * from './lib/AccessoryLoader';
 export * from './lib/StreamController';
 export * from './lib/HAPServer';
 export * from './lib/gen';
+export * from './lib/datastream';
 
 export * from './lib/util/chacha20poly1305';
 export * from './lib/util/clone';
--- a/src/lib/datastream/DataStreamManagement.ts
+++ b/src/lib/datastream/DataStreamManagement.ts
@@ -1,0 +1,230 @@
+import * as tlv from '../util/tlv';
+import bufferShim from "buffer-shims";
+import createDebug from "debug";
+import {Service} from "../Service";
+import {
+    Characteristic,
+    CharacteristicEventTypes,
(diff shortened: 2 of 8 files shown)


Supereg:
rebased onto master

Supereg:
rebased onto master (again)

KhaosT:
Looks really cool ^^ Regarding `int64` problem, if we really want, maybe expose those as `Decimal` using the bundled `decimal.js`?

KhaosT:
And thanks for implementing HDS 🙏

PR #739, opened 2019-11-09
Adding support for Siri remotes

Supereg:
This part two of two (with part one being #736 and thus also containing the changes of the previous PR).

In this PR support for Siri voice input has been added. The provided example plugin was extended to show how supporting voice data is going to work (by default turned off). The only thing which you need to define is audio producer object, which generates individual opus frames (according to the supplied audio configuration) and calls the respective handler to send them out. I decided to include an example implementation which uses gstreamer and `alsasrc` (as most of the people run hap-nodejs on raspberry pis anyway. A implementation using `osxaudiosrc` would also be possible. However I couldn't get gstreamer to record any audio under macOS Catalina when executing gstreamer directly via terminal).

Also in this PR:
* The remote controller is generally more robust and fixed some issues (most importantly stuff related to updating button configurations)
* Additionally added some events to the remote controller class so it can be properly integrated into UI, devices, etc
* Added an `PAIRED` and `UNPAIRED` event to the Accessory class (`UNPAIRED` is the important one as it is used to reset the target configuration of the remote)

Note: For homebridge users the current implementation is a bit limiting. It requires an implementation via a homebridge platform plugin, as the `addServicesToAccessory(...)` method requires a reference to the accessory in order to correctly configure the services and events. This is problematic since simple homebridge plugins do not have a reference to the accessory instance.

### DIFF
--- a/src/accessories/AppleTVRemote_accessory.ts
+++ b/src/accessories/AppleTVRemote_accessory.ts
@@ -1,10 +1,9 @@
-import { Accessory, Categories, uuid } from '..';
-import { ButtonState, ButtonType, HomeKitRemoteController } from "../lib/HomeKitRemoteController";
+import { Accessory, ButtonState, ButtonType, Categories, HomeKitRemoteController, uuid } from '..';
 import * as http from "http";
 import url, { UrlWithParsedQuery } from "url";
+import { GStreamerAudioProducer, GStreamerOptions } from "./gstreamer-audioProducer";
 
 const remoteUUID = uuid.generate('hap-nodejs:accessories:remote');
-
 const remote = exports.accessory = new Accessory('Remote', remoteUUID);
 
 // @ts-ignore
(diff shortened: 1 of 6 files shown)


KhaosT:
Can you move `gstreamer-audioProducer.ts` from `lib` to `accessories` since it's used by the example of the accessory implementation, instead of the core library?

Siri Remote profile always felt a really weird thing for Apple to do... 😝

Supereg:
rebased onto master

Supereg:
> Can you move `gstreamer-audioProducer.ts` from `lib` to `accessories` since it's used by the example of the accessory implementation, instead of the core library?

The main idea was to provide everything required to run a minimal setup for Siri remotes. As most people should be fine running the supplied gstreamer/alsa suite and I wanted to avoid fragmentation resulting by people copying the same code into their project. As far as I know the files in `accessories` are not exported (?). But on the other hand I do understand your reasoning behind the idea of moving it into `accessories`.

KhaosT:
Yeah, I consider gstreamer as a "specific" implementation of Siri Remote, and depends on the context (like what accessory people try to use this for), gstreamer might not fitting their needs.

PR #742, opened 2019-11-10
Bug fixes

codyc1515:
### Changes
- Fixes warning about Accessory Name not matching
- Fixes Bridge characteristic Version permissions
- Fixes sample accessory Thermostat
- Fixes TV characteristic Picture Mode format

I could do with some help on refactoring the Accessory Name code, but it works to solve the warning message from occurring in the HomeKit logs and complies with the spec.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -187,7 +187,7 @@ export class Accessory extends EventEmitter<Events> {
     // create our initial "Accessory Information" Service that all Accessories are expected to have
     this
       .addService(Service.AccessoryInformation)
-      .setCharacteristic(Characteristic.Name, displayName)
+      .setCharacteristic(Characteristic.Name, displayName + " " + crypto.createHash('sha512').update(UUID, 'utf8').digest('hex').slice(0, 4).toUpperCase())
       .setCharacteristic(Characteristic.Manufacturer, "Default-Manufacturer")
       .setCharacteristic(Characteristic.Model, "Default-Model")
       .setCharacteristic(Characteristic.SerialNumber, "Default-SerialNumber")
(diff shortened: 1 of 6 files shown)


src/lib/Accessory.ts
```diff
@@ -187,7 +187,7 @@ export class Accessory extends EventEmitter<Events> {
     // create our initial "Accessory Information" Service that all Accessories are expected to have
     this
       .addService(Service.AccessoryInformation)
-      .setCharacteristic(Characteristic.Name, displayName)
+      .setCharacteristic(Characteristic.Name, displayName + " " + crypto.createHash('sha512').update(UUID, 'utf8').digest('hex').slice(0, 4).toUpperCase())
```
KhaosT: Instead of updating this at multiple places, let's just rewrite `displayName` saved on the instance and use that across places maybe?

src/lib/Accessory.ts
```diff
@@ -225,6 +225,20 @@ export class Accessory extends EventEmitter<Events> {
...
+    if(typeof this.username !== 'undefined') {
+        // @ts-ignore
+        if(typeof service.displayName !== 'undefined') {
+            // @ts-ignore
+            service.displayName = service.displayName + " " + crypto.createHash('sha512').update(this.UUID, 'utf8').digest('hex').slice(0, 4).toUpperCase();
+        }
+        // @ts-ignore
+        if(typeof service.characteristics[0].displayName !== 'undefined' && service.characteristics[0].displayName == "Name") {
+            // @ts-ignore
+            service.characteristics[0].value = service.characteristics[0].value + " " + crypto.createHash('sha512').update(this.UUID, 'utf8').digest('hex').slice(0, 4).toUpperCase();
```
KhaosT: We probably should not rewrite service name? Is there a particular reason why we want to do that?

src/lib/Accessory.ts
```diff
@@ -630,7 +644,7 @@ export class Accessory extends EventEmitter<Events> {
     this._accessoryInfo.setupID = this._setupID;
 
     // make sure we have up-to-date values in AccessoryInfo, then save it in case they changed (or if we just created it)
-    this._accessoryInfo.displayName = this.displayName;
+    this._accessoryInfo.displayName = this.displayName + " " + crypto.createHash('sha512').update(this.UUID, 'utf8').digest('hex').slice(0, 4).toUpperCase();
```
KhaosT: If we rewrite `this.displayName` then we don't need to do anything special here ^^

codyc1515:
I think that changing the naming to be consistent throughout is probably going to be too challenging for me to do properly, so I'm going to revert this but would be good to have it fixed at some stage. Right now it does not cause any issues.

Can we please get the other bug fixes included?

KhaosT:
Sure ^^

KhaosT:
Thanks for fixing the other issues ^^ it’s interesting that they changed format of picture mode after shipping iOS with uint16 😅

PR #747, opened 2019-11-10
Use short service and characteristic type UUIDs

### LINKED ISSUE #740: Code does not comply with 5.6.1 Service and Characteristic Types of the HAP spec
I would fix this, but am not sure how to. In practise, I don't think this breaks anything but it means that we are not compliant with the spec. Relevant extract from the public spec:

### 5.6.1 Service and Characteristic Types
When Apple-defined UUIDs are encoded as JSON strings, a short form must be used by including only the first 8 characters with leading zeros removed. When converting short form UUIDs back to full UUIDs, the process is reversed by prefixing the string with 0 or more '0' characters to expand it to 8 characters then "-0000-1000-8000-0026BB765291" is appended to form the full 36-character UUID.

For example, the public.hap.service.accessory-information full UUID is "0000003E-0000-1000-8000-0026BB765291". When encoded as a JSON string, the first 8 characters are "0000003E" and with leading zeros removed it becomes "3E". Converting back to a full UUID would prefix "3E" with six '0' characters to get "0000003E" then append "-0000-1000-8000-0026BB765291" to form a full UUID of "0000003E-0000-1000-8000-0026BB765291". Other examples:

samuelthomas2774:
- Respond with short type UUIDs for Apple-defined services and characteristics (fixes #740)
- Require a description and type UUID for characteristics (these are currently optional?)
- Fix service.getCharacteristic not automatically adding the Name characteristic


codyc1515:
This causes the warning to go away for short form Apple UUIDs. Happy to approve.

Supereg:
> This causes the warning to go away for short-form Apple UUIDs.

Warnings appearing in the console app or what warnings are you talking about?

src/lib/Service.ts
```diff
@@ -251,11 +256,11 @@ export class Service extends EventEmitter<Events> {
           return this.addCharacteristic(name);
         }
       }
-      //Not found in optional Characteristics. Adding anyway, but warning about it if it isn't the Name.
+      // Not found in optional Characteristics. Adding anyway, but warning about it if it isn't the Name.
       if (name.UUID !== Characteristic.Name.UUID) {
         console.warn("HAP Warning: Characteristic %s not in required or optional characteristics for service %s. Adding anyway.", name.UUID, this.UUID);
-        return this.addCharacteristic(name);
       }
+      return this.addCharacteristic(name);
```
Supereg, line 263: Could someone elaborate on this change? I don't know the history of not adding the Name characteristic and don't really know if that affects anything?

samuelthomas2774: This fixes the Name characteristic not being added by `getCharacteristic`, which shouldn't affect most users anyway as the Service constructor adds the display name if it's set there.

Supereg: Okay checked the git history, doesn't seem to have any reason behind that.

PR #748, opened 2019-11-11
Add categories to accessories

codyc1515:
Does not seem to work on the Thermostat for some reason, but works on all others.


codyc1515:
Re-based @KhaosT

PR #751, opened 2019-11-11
Fix Thermostat accessory formats and permissions

codyc1515:


src/lib/gen/HomeKit-TV.ts
```diff
@@ -15,7 +15,7 @@ export class ActiveIdentifier extends Characteristic {
     super('Active Identifier', ActiveIdentifier.UUID);
     this.setProps({
       format: Formats.UINT32,
-      perms: [Perms.READ, Perms.WRITE, Perms.NOTIFY]
```
KhaosT: This actually needs to be `read, write, notify`, as HomeKit TV profile writes this characteristic when user switch input.

codyc1515: Apologies, you are right. It is only on the Target Control service that Active Identifier is not meant to have Write permissions. I'm not entirely sure how to fix this.

KhaosT: @codyc1515 interesting, is homed rejecting the accessory with target control service in that case?

codyc1515: No it is not. However, I just checked the spec again & it is 100% only meant to be paired read & notify. What functionality specifically does this break?

KhaosT: @codyc1515 Does the spec you checked has TV profile on it? This change happened with TV profile, you can verify the latest expected permission from iOS via metadata in homed.

Particularly for this one, if we remove `write` permission, people using HomeKit TV will no longer be able to change input, since homed will reject write 😅

codyc1515: Hmm in that case if it is not breaking anything I will revert it.

KhaosT:
🙏 Thank you

Issue #734, opened 2019-11-03
tsc --build tsconfig.json fails


ilendemli:
I tried to compile using tsc because the project runs awfully slow on a Raspberry Pi Zero W since the TypeScript merge but it fails with the following error:

```
pi@raspberrypi:~/HAP-NodeJS $  tsc --build tsconfig.json
src/lib/gen/importAsClasses.ts:182:9 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'typeof Formats'.
  No index signature with a parameter of type 'string' was found on type 'typeof Formats'.

182     if (Formats[key] == format)
            ~~~~~~~~~~~~

src/lib/gen/importAsClasses.ts:191:9 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'typeof Units'.
  No index signature with a parameter of type 'string' was found on type 'typeof Units'.

191     if (Units[key] == units)
            ~~~~~~~~~~


Found 2 errors.
```

Shaquu:
Well it looks like you have some rules in tsconfig.json that cause an error and maybe it is not important for the whole project to work properly?

Can you share what is inside of yours tsconfig.json?

ilendemli:
@Shaquu it's the same [tsconfig.json](https://github.com/KhaosT/HAP-NodeJS/blob/master/tsconfig.json) one as in this project, running it with ts-node is no problem, it just takes ages to start and parse accessories.
I wanted to compile and try it that way but encountered that error.

gkaindl:
The problem is that the both ```Formats``` and ```Units``` are enums with string values now, and TypeScript complains about accessing the values directly with string keys.

You can change the offending lines to

```if (Formats[key as keyof typeof Formats] == format)```

and

```if (Units[key as keyof typeof Units] == units)```

to fix this.

I'm unsure if someone should make a pull request for a fix like this: The offendign file is in the ```/gen``` subdirectory, which sounds like it's being generated from something, but I can't find the the script generating ```importAsClasses.ts```. It would make more sense to fix it there, rather than in the generated code.

PR #753, opened 2019-11-13
Add missing type cast to fix TypeScript build.

gkaindl:
Currently, building via `tsc` fails due to two missing type casts in `importAsClasses.ts` -> this doesn't affect running via `ts-node`, because the affected code is apparently not used anywhere else.

However, since the switch to TypeScript, building before running is quite advisable on embedded platforms (RPI et al.), since `ts-node` is incredibly slow there.

This pull requests adds the missing casts, so that `tsc` finishes without errors.

Also see #734


codyc1515:
Thank you for this. I am running a Raspberry Pi 4 and this is substantially quicker than using the ts-node command (virtually immediate).

PR #756, opened 2019-11-23
Bug fixes for Secure Video, Wi-Fi Router & permissions

codyc1515:


src/lib/gen/HomeKit-TV.ts
```diff
@@ -15,7 +15,7 @@ export class ActiveIdentifier extends Characteristic {
     super('Active Identifier', ActiveIdentifier.UUID);
     this.setProps({
       format: Formats.UINT32,
-      perms: [Perms.READ, Perms.WRITE, Perms.NOTIFY]
```
KhaosT: 👋 It really needs the write permission...

codyc1515:
Have reverted this. I think I made the same mistake last time. I don't know why the spec doesn't reflect this.

Supereg:
Out of interest, what's the ProductData characteristic for?

Supereg:
Additionally, my resources and informations suggest different values for most changes you suggested. Could you specify where you got your information from?

KhaosT:
@Supereg Those matched Logi Circle 2's implementation I think? https://github.com/KhaosT/HAP-NodeJS/issues/741

PR #759, opened 2019-11-29
Fix unpublish function to close all open connections and allow republish

gotofoo:
After calling unpublish the HTTP server will be stopped but all clients stay connected which you can see if you call netstat. Any data which will be sent from those connected clients will still be processed even though the bridge should not be published anymore at this point. The process must be killed to close those open connections.

This makes it impossible to unpublish and republish a bridge. Furthermore the publish method cannot be called again because the ProtocolInformation service will be added again. The application needs to be restarted in order to be able to publish the bridge again.

Those issues lead to the issue described under https://github.com/NRCHKB/node-red-contrib-homekit-bridged/issues/12 where it is not possible to redeploy a bridge node in Node-RED by calling unpublish and creating a new bridge instance.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -623,8 +623,13 @@ export class Accessory extends EventEmitter<Events> {
    *                                new Accessory.
    */
   publish = (info: PublishInfo, allowInsecureRequest?: boolean) => {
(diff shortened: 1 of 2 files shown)


Shaquu:
Well, that’s hot.

KhaosT:
🙏 Thanks for fixing this

PR #763, opened 2019-12-16
Increase type strictness, disable esModuleInterop

meyer:
This PR contains a few changes, all in the interest of stronger typing and greater compatibility for `hap-nodejs` as a library:

- I disabled `esModuleInterop` and converted implicit namespace imports to explicit namespace imports. This is potentially controversial but for libraries it’s generally the better way to go. Libraries with `esModuleInterop` enabled require downstream consumers to enable `esModuleInterop` in their projects, whether or not they’d like to. `esModuleInterop` oddly makes libraries _less_ interoperable.
- I enabled `noUnusedLocals` and then removed, commented out, or exported unused variables and functions. Feel free to nit-pick with what I changed—I’ve got no opinions other than "unused code shouldn’t be left lying around".
- I updated TypeScript from 3.5.3 to 3.7.3 to be able to use the optional chaining feature ("the question mark") that was added in 3.7. Since your compile target is still `"es5"` the output code will remain ES5-compatible.
- I fixed some `@ts-ignore` comments (this is where I used optional chaining).
- I fixed some type issues (`private constructor`).
- I removed the yarn lockfile since it appears that `npm` is the preferred package manager.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -474,6 +474,22 @@
       "integrity": "sha512-pMCcqU2zT4TjqYFrWtYHKal7Sl30Ims6ulZ4UFXxI4xbtQqK/qqKwkDoBFCfooRqqmRu9vY3xaJRwxSh673aYg==",
       "dev": true
     },
+    "@types/node-persist": {
+      "version": "0.0.33",
+      "resolved": "https://registry.npmjs.org/@types/node-persist/-/node-persist-0.0.33.tgz",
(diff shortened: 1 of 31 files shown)


Supereg:
Type strictness was already improved in recent version (and there is still the goal to move to eslint in the near future).
There is currently no plan to disable esModuleInterop.
Thus I will close this PR (and it's terribly outdated).

meyer:
@Supereg thanks for the update

PR #764, opened 2019-12-17
Install, configure, and run prettier

meyer:
While I was digging around this codebase for #763 I noticed that there were a few different code formatting styles in use. For ease of future contributions I installed prettier and configured it to match the current code style as closely as possible. I also updated the `Node-CI` GitHub action to run prettier with the `--check` flag to ensure that prettier has been run on subsequent commits.

To format, you can either configure your editor to format with prettier on save or you can run `npm run format` to reformat code in the `src` folder.

The prettier config (in `package.json`) could do with some opinions. Feel free to modify it or request changes!

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -21,7 +21,8 @@ jobs:
       - name: npm install, build and test
         run: |
           npm ci
-          npm run build --if-present
+          npm run format-check
+          npm run build
           npm test
         env:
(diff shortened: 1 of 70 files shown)


Supereg:
There are currently no plans for this project to start adopting prettier.
However there are plans to move to ESLint in the near future in the same way as it was already introduced in the latest beta of the homebridge project.

Issue #695, opened 2019-08-03
Server not listening on IPv4


illiteratewriter:
In `util/eventedhttp.js`, `EventedHttpServer.prototype.listen`  you can only pass targetPort and not the host. This works in most of the scenarios, since 

> In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

But, in Ubuntu the server is listening only on the IPv6. It would be great if there can be some way to add the host (which is the second parameter in `server.listen([port[, host[, backlog]]][, callback])` ) so that we can configure if needed for the server to listen to IPv4. One example use case would be for this https://github.com/nfarina/homebridge/issues/1277 issue.

buckmelanoma:
Still an issue for me.  I had to manually edit eventedhttp.js to restore IPv4

Supereg:
The latest beta adds an option to specify a `PublishInfo` option to specify on which interfaces to listen for, as well as limiting to ipv4 only.

PR #766, opened 2019-12-18
Attach to interface

Roman1us:
Allow set host for HAP server #695

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -86,7 +86,7 @@ export enum AccessoryEventTypes {
 
 type Events = {
   identify: (paired:boolean, cb: VoidCallback) => void;
-  listening: (port: number) => void;
+  listening: (port: number, host: string) => void;
   "service-configurationChange": VoidCallback;
   "service-characteristic-change": (change: ServiceCharacteristicChange) => void;
   [AccessoryEventTypes.PAIRED]: () => void;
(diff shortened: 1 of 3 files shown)


KhaosT:
Should we also update Advertiser to only advertise on the given interface?

Supereg:
> Should we also update Advertiser to only advertise on the given interface?

Would be relatively important. @Roman1us could you add that?

Roman1us:
> > Should we also update Advertiser to only advertise on the given interface?
> 
> Would be relatively important. @Roman1us could you add that?

Thinking about it, but we have a mdns object in configuration. In tests with mDNSResponder and avahi-daemon run in system i've got problems with attaching to interfaces that already use this interface

Roman1us:
Have somthing to discuss?

Supereg:
Will probably ship with v0.8.0, needs some work in the underlying mdns library

Roman1us:
> Will probably ship with v0.8.0, needs some work in the underlying mdns library

Ok, i'm going to test ciao library in different environments

Supereg:
alright, so with a little bit of delay, the feature was finally implemented and will ship with the next release (follow [release cycle](https://github.com/homebridge/HAP-NodeJS/projects/4)).

The feature was implemented in a separate commit, as it is a bit more sophisticated and provides much greater control (and as it required additional tailoring to properly change the interface in the advertising library). So the PR will just be closed.
Nonetheless thanks for your contribution, as this was the fundamentally necessary starter for this change.

The new `bind` publish option, can be a string or array of strings containing ip addresses or network interface names.
For the best overview on how it works, you may just read the documentation linked below.
The feature is currently still in beta. Feel free to test if you want.

https://github.com/homebridge/HAP-NodeJS/blob/e858e39576d1b4282d477b78d0e432f417721a03/src/lib/Accessory.ts#L156-L203

PR #768, opened 2019-12-22
Automatically update configuration after specific characteristics change

gotofoo:
While I was working on the issue https://github.com/NRCHKB/node-red-contrib-homekit-bridged/issues/12 I noticed that the clients don't get informed about changes of some core characteristics (e.g. Name, Firmware Revision, Model, ...). To inform the clients about those kind of changes the configuration version (TXT Record key "c#") needs to be increased to tell the clients that the configuration has changed.

To automatically update the configuration after specific characteristics changed I defined a new property for characteristics. Every time a characteristic gets changed which has this property set to true the configuration will automatically be updated. This property defaults to false and can be overwritten in the characteristic definition.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -267,6 +267,16 @@ export class Accessory extends EventEmitter<Events> {
       if (!this.bridged)
         this._handleCharacteristicChange(clone(change, {accessory:this, service:service as Service}));
 
+      // if the characteristic is configured to cause a config update, then
+      // the configuration of the bridge will be updated which will inform the
+      // clients about the change
+      if (change.characteristic.props.configUpdate) {
+        if (this.bridge) {
+          this.bridge._updateConfiguration();
+        } else {
+          this._updateConfiguration();
+        }
+      }
     });
 
     return service;
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -67,6 +67,7 @@ export interface CharacteristicProps {
   maxDataLen?: number;
   validValues?: number[];
   validValueRanges?: [number, number];
+  configUpdate?: boolean;
 }
 
 export enum Access {
@@ -361,6 +362,7 @@ export class Characteristic extends EventEmitter<Events> {
       minValue: null,
       maxValue: null,
       minStep: null,
+      configUpdate: false, // default to no update
       perms: []
     };
   }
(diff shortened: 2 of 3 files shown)


Shaquu:
@KhaosT please look at this PR.

KhaosT:
Based on HomeKit Accessory Protocol Specification, those information are not suppose to change during the lifetime of the accessory. Since homed does the special parsing for the accessory info service, I'm not sure we want to support this use case since there were assumptions made to make this work (homed will rediscover accessory after configuration number changes). Is there a particular reason that those information can't be populated when you add the accessory?

joconor:
One case I'm aware of is https://github.com/naofireblade/homebridge-weather-plus (a Homebridge plugin) dynamically changes the Name characteristic of some Accessories. It will create "Occupancy Sensor" Accessories to reflect when a condition exceeds some threshold (e.g., Wind Speed) and set the sensor as "Home". The Occupancy Sensor Value reflects whether or not the threshold has been exceeded, but the Name Characteristic is changed dynamically to reflect the current value (e.g., the current wind speed).
The Apple Home app will occasionally re-read the characteristics (I'm not quite sure under what conditions. New session, maybe?) and the Name will reflect the change at that time, but there's currently no way for an Accessory to 'push' a Name change other than incrementing the configuration count.

gotofoo:
@KhaosT I missed that part of the specification were this was mentioned but you are right. It states that the values of Manufacturer, Model, Name and Serial Number should persist throughout the lifetime of an accessory. This has probably some compliance reasons and not so much technical reasons. Then it is probably better not to implement this change even though that it seams to be working just fine.

To really follow the specifications it should probably not be allowed to change those specific characteristics at all after the accessory was published the first time since it isn't really reliable. Instead this should lead to an error or a warning message. This means that a new accessory must be created every time someone wants to change those fields.

This limitation should be documented somewhere in the Wiki section.

Supereg:
@joconor Are you sure the plugin modifies the Name characteristic of the Accessory Information service? Or does it modify the Name characteristic of the individual services (Occupancy sensor etc). These are two fundamentally different things. The second one does perfectly work as far as I know(?)

KhaosT:
Yeah, it seems in general changing accessory information service is always problematic unless configuration number is bumped along with that... I'll close this one for now ^^

PR #778, opened 2020-01-16
Remove obsolete dependency of buffer-shims

Supereg:
This PR removes the dependency of `buffer-shims` as node v4 support was dropped.


KhaosT:
Sweet :)

PR #780, opened 2020-01-18
Adding serialization and deserialization functions for Accessory, Services and Characteristics

Supereg:
Added for serialization and deserialization functions Accessory, Services and Characteristics to be used in homebridge as mentioned here in [homebridge/2391](https://github.com/nfarina/homebridge/pull/2391#issuecomment-575771704).

The goal is to improve maintainability and stability, so homebridge does not need to work on hap-nodejs internals.

Additionally the following things are now serialized correctly:
* Characteristic.accessRestrictedToAdmins
* Service.isHiddenService
* Service.isPrimaryService
* Service.optionalCharacteristics

### DIFF
--- a/src/lib/Service.ts
+++ b/src/lib/Service.ts
@@ -1,6 +1,7 @@
 import {
   Characteristic,
-  CharacteristicEventTypes
+  CharacteristicEventTypes,
+  SerializedCharacteristic
 } from './Characteristic';
 import { clone } from './util/clone';
 import { EventEmitter } from './EventEmitter';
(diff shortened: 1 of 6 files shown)


Supereg:
Just have seen, my editor rearranged some imports, gonna revert those unnecessary changes tomorrow.
Additionally some automated test could be beneficial to implement.

Supereg:
Fixed import formatting and added some test (which was a good idea to add 😅 fixed some 'little' issue in deserialization in Accessory).

Additionally the Advertiser test was failing. I fixed that if it's okay.

Before you release this changes under a new version, I would like to submit an additional PR which introduces some more stuff which I plan to use in the homebridge update (so no need to hurry; this notice is only just in case :))

src/lib/Service.ts
```diff
@@ -364,4 +377,35 @@ export class Service extends EventEmitter<Events> {
...
+    return {
+      displayName: service.displayName,
+      UUID: service.UUID,
+      subtype: service.subtype,
+
+      hiddenService: service.isHiddenService,
+      primaryService: service.isPrimaryService,
+
+      characteristics: service.characteristics.map(characteristic => Characteristic.serialize(characteristic)),
+      optionalCharacteristics: service.optionalCharacteristics.map(characteristic => Characteristic.serialize(characteristic)),
```
KhaosT: I'm not sure we need/want to serialize `optionalCharacteristics`? If developer expressed intent to use that characteristic, it will be added to `characteristics` when the first time they initialize the service...

Supereg: Yeah definitely not necessary (that's probably why homebridge never implemented it).
My intention was to fully restore a characteristic. If we don't restore optionalCharacteristics hap-nodejs will [complain](https://github.com/KhaosT/HAP-NodeJS/blob/1d60575b9e59327702c208d6f9e2d175f8c57a55/src/lib/Service.ts#L250) about optional characteristics being added (when a plugin dev later decides to upgrade his services). Already cached accessories can't be changed anymore though.

But I can also remove serialization of them again 🤷🏽‍♂️

KhaosT:
Okay ^^

PR #781, opened 2020-01-18
Adding configuration for GitHub actions CI

Supereg:
This PR is a proposal to add a CI runner to the project.
There is no real benefit, if we have some unit tests but never run them and regressions potential get published unnoticed.

Additionally I fixed an already broken Tests in this PR.

I don't really care what CI runner we are going to use, I just picked Travis, since the integration with GitHub seems pretty good and it worked for me in the past.
Additionally you could hook up travis to [deploy npm releases](https://docs.travis-ci.com/user/deployment/npm/) (a new version is published for every tagged commit). For me this feature wasn't always reliable, but in every case you would need to finish the setup @KhaosT (setup travis for your account and maybe add deployment tokens).

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,27 @@
+name: Node CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 13.x]
+        os: [ubuntu-latest, macOS-latest]
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: npm install, build and test
+        run: |
+          npm ci
+          npm run build --if-present
+          npm test
+        env:
+          CI: true
--- a/src/lib/Advertiser.spec.ts
+++ b/src/lib/Advertiser.spec.ts
@@ -3,7 +3,7 @@ import { AccessoryInfo } from './model/AccessoryInfo';
 import './gen';
 
 const createAdvertiser = () => {
-  return new Advertiser(new AccessoryInfo('displayName'), {
+  return new Advertiser(AccessoryInfo.create('00:00:00:00:00:00'), {
     multicast: false,
     interface: 'ipv6',
     port: 80,


KhaosT:
Just curious, have you thought about using GitHub Actions? Seems pretty cool and easier to integrate?

Supereg:
Not really. If the same can be achieved would probably be better to have everything in one place.
Checking the Actions Tab, I would assume only you could configure those?

Supereg:
Just checked, there are some predefined workflows for NPM.
This will create a new yml at `.github/workflows/nodejs.yml` with the following default content
```
name: Node CI

on: [push]

jobs:
  build:

    runs-on: ubuntu-latest

    strategy:
      matrix:
        node-version: [8.x, 10.x, 12.x]

    steps:
    - uses: actions/checkout@v1
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - name: npm install, build, and test
      run: |
        npm ci
        npm run build --if-present
        npm test
      env:
        CI: true
```

Only thing what can be changed should be on to include `pull_request` and remove node-version 8.x and add maybe 13.x

I could change this PR to include this file or you just create the workflow over the actions tab. If we want to use GitHub actions at all.

KhaosT:
Do you have a strong preference going with Travis CI? If not, I can get GitHub Actions configured today ^^

Supereg:
Currently playing around a bit with GitHub actions. And I got now a strong preference to go with GitHub Actions 😂 it's a bit more verbose regarding the configuration but it seems also quite power flow and pretty well integrated.

Supereg:
Added a example config were we also test on any supported platform. Maybe a bit too much 😅 but would be cool if you set up a workflow :)

Additionally I think there is also a 'Node.js Package' workflow for CD which runs on every new release

Supereg:
Okay looking at my test GitHub action it seems like our tests won't run on windows (wasn't windows a supported platform? 😅)

Supereg:
It seems like jest doesn't run at all on windows instances? https://github.com/facebook/jest/issues/9413

Supereg:
Okay, got now a good overview over GitHub Actions. Windows as os can currently not be used as jest just crashes with exist code 1.
Fail-fast is currently on. So when any job fails early all others get cancelled automatically.

Lastly, I think GitHub Actions could get pretty costly:
Bildschirmfoto 2020-01-18 um 20 32 24

This was just me testing some stuff. Don't know how much HAP-NodeJS would consume. We could potentially reduce the amount of node versions tested and exclude macOS as it calculates 10x the minutes for macOS runs lol

KhaosT:
@Supereg GitHub Actions is free for public repo ^^

KhaosT:
@Supereg can you see the run? https://github.com/KhaosT/HAP-NodeJS/runs/397037994

Supereg:
> @Supereg GitHub Actions is free for public repo ^^
Oh lol, must have confused something 

> @Supereg can you see the run?
Yes ☺️👏🏽

PR #784, opened 2020-01-19
Adding CD using GitHub Actions

Supereg:
Added a second job which automatically publishes the npm package.

It does only run if all previous build jobs were successful, the triggering event was `push` (so it doesn't run for pull_request or other future events) and the job got executed because a new version tag (a tag that starts with v; we cannot really use regex to match the version string exactly) was pushed.

For this to work you would need to add a secret called `npm_token` to the project (via settings) which contains a npm publish token.

Edit: You could also create a completely new workflow for publishing which only runs on the 'release' event. However you would then need to manually create a new release in GitHub to trigger this event. As this project didn't create a single GitHub release, I think checking for tags is probably the better way.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -25,3 +25,22 @@ jobs:
           npm test
         env:
           CI: true
+
+  publish-npm:
+    # publish should only ran on 'push' event and if a version tag was created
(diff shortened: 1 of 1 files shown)


donavanbecker:
@Supereg Just implemented this on HomeBridge.

PR #786, opened 2020-01-20
Improve setPrimaryService API a bit

Supereg:
Simplified the setPrimaryService API call a bit.

`setPrimaryService` is now a member of the Service class (the old method in the Accessory class got deprecated and just forwards the call to this method). So now primary, hidden and linked service management is all at the same place. This enables homebridge plugin devs to set primary services (wasn't really possible before).

Removing primary services and adding them again could lead to multiple primary services. This is now taken care of correctly.

### DIFF
--- a/src/accessories/AirConditioner_accessory.ts
+++ b/src/accessories/AirConditioner_accessory.ts
@@ -92,6 +92,7 @@ FanService.addCharacteristic(Characteristic.RotationSpeed)
 
 var ThermostatService = ACTest.addService(Service.Thermostat,"Thermostat");
 ThermostatService.addLinkedService(FanService);
+ThermostatService.setPrimaryService();
 
 ThermostatService.getCharacteristic(Characteristic.CurrentHeatingCoolingState)!
 
@@ -156,5 +157,4 @@ LightService.getCharacteristic(Characteristic.On)!
       console.log( "Characteristic Light On changed to %s",value);
       callback();
     });
-LightService.setHiddenService(true);
-ACTest.setPrimaryService(ThermostatService);
+LightService.setHiddenService();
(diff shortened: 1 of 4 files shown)


KhaosT:
I'm not sure this change make sense? The idea of "primary service" exists on the accessory level. Without an accessory, a service marked as primary doesn't mean anything, and exposing an API on service object for this purpose seems... less ideal.

Supereg:
As already noted in the other thread, enjoy your holiday and do not feel forced to reply. I got time ☺️

------

Well that's a fair point.
However being primary or not I still a property represented by the service itself. So it definitely, somehow belongs to a service(?). The service **is** the one who is primary, an accessory is the one who **has** a primary service. So it kind of belongs to a service but only really means anything if the service get's added to an accessory?
Maybe you could also argue, that when you built an array of services (like literally Service[]), having the proposed API you could mark the primary accessory in this array. Meaning, maybe the idea of a "primary service" is not at all a thing which is represented by an accessory. Rather this idea is more generally represented by a set of services. Accessory just happens to be another way to represent a set of services? Meaning an `Accessory` as well as a plain `Service[]` can have a primary service.
So looking at it that way, I think you could probably argue for both sides?

The though behind the proposal of this API change was mainly inspired by my ambitions of abstracting the controller API (which is currently pretty semi consistent). Just like a currently existent CameraSource a Controller (on the new abstraction level) would just return a set of services (Service[]). As explained above, this API change would allow marking the primary service in this set of services.
This PR originated in the idea of trying to not give controllers access to the accessory at all. This change would allow that.

A nice side effect is that without any homebridge specific workarounds, all homebridge plugins (so platforms **as well as** accessories) are now able to mark primary services.

---------

As I already stated, you can probably find good arguments for both side. I originally didn't want to argue on that kind of philosophical level you only really find in Software engineering theory classes 😂😂. Ultimately you are the one who decides what get's merged and what not. I'm honestly fine with both.

PR #793, opened 2020-02-15
Ability to catch 'EADDRINUSE'

marcus-j-davies:
So,
This will be my first ever PR on github (I'm kind of the new kid around here) - Be Gentle.

I use HAP-NodeJS in a non-bridged manner, in essence, I have created a collection of nodes for Node-Red, that contains (currently) 10 pre-configured, fully functional devices, including most of the optional characteristics, to make them as fully featured & ready to use as possible. 

After much studying (and of course testing), I understand we can be notified, when an Accessory has instantiated, by subscribing to `LISTENING`.

However, if one were to not change the listening port (Nodes can be copied/pasted retaining its port assignments), all hell breaks loose, and you end up with a crashed Node-Red environment (or any other host potentially that HAP is running under).

My modifications are to introduce an `ADDRESS_IN_USE ` event, that bubbles up through to the Accessory object.

What we end up with, is either the `LISTENING` or `ADDRESS_IN_USE` event being raised, with no crashed environment, allowing the developer to handle it, in much better way.

Testing this approach seems to have the desired affect.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -90,6 +90,7 @@ export enum AccessoryEventTypes {
   SERVICE_CHARACTERISTIC_CHANGE = "service-characteristic-change",
   PAIRED = "paired",
   UNPAIRED = "unpaired",
+  ADDRESS_IN_USE = "address-in-use"
 }
 
 type Events = {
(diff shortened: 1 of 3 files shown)


KhaosT:
Hi, thanks for the PR. Is there a particular reason you can’t just let the accessory automatically assign a port? If you invoke accessory.publish without a port, the library will automatically find a port that’s free and use that…

marcus-j-davies:
You learn something new everyday, all examples I have seen, include specifying the port, so I was not aware of that ability.

I guess, if one were to specify a port, then copy and paste that same config - the issue could still occur?

PR #798, opened 2020-03-31
Adding new SmartSpeaker service

Supereg:
iOS 13.4 seems to have added a new SmartSpeaker service.

MediaStates kind of work in the Home App (when switching between playing and paused the loading indicator will stay though).
Volume and Mute Characteristics seem to not have any effect.

It seems a device manufacturer would need to implement AirPlay 2 additionally to have a fully working HomeKit Speaker:
IMG_3279

### DIFF
--- a/src/accessories/SmartSpeaker_accessory.ts
+++ b/src/accessories/SmartSpeaker_accessory.ts
@@ -1,0 +1,61 @@
+import {
+    Accessory,
+    Categories,
+    Characteristic,
+    CharacteristicEventTypes,
+    CharacteristicGetCallback, CharacteristicSetCallback, CharacteristicValue,
+    Service,
+    uuid
+} from "..";
+import {CurrentMediaState, TargetMediaState} from "../lib/gen/HomeKit-TV";
+
+const speakerUUID = uuid.generate('hap-nodejs:accessories:smart-speaker');
+const speaker = exports.accessory = new Accessory('SmartSpeaker', speakerUUID);
+
+// @ts-ignore
+speaker.username = "89:A8:E4:1E:95:EE";
+// @ts-ignore
+speaker.pincode = "676-54-344";
+speaker.category = Categories.SPEAKER;
+
+const service = new Service.SmartSpeaker('Smart Speaker', '');
+
+let currentMediaState: number = CurrentMediaState.PAUSE;
+let targetMediaState: number = TargetMediaState.PAUSE;
+
+// ConfigureName is used to listen for Name changes inside the Home App.
+// A device manufacturer would probably need to adjust the name of the device in the AirPlay 2 protocol (or something)
+service.setCharacteristic(Characteristic.ConfiguredName, "Smart Speaker");
+service.setCharacteristic(Characteristic.Mute, false);
+service.setCharacteristic(Characteristic.Volume, 100);
+
+service.getCharacteristic(Characteristic.CurrentMediaState)!
+    .on(CharacteristicEventTypes.GET, (callback: CharacteristicGetCallback) => {
+        console.log("Reading CurrentMediaState: " + currentMediaState);
+        callback(undefined, currentMediaState);
+    }).getValue();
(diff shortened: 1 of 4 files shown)


Supereg:
Would be cool if someone could verify that indeed that service was not present in version prior to iOS 13.4 (and related version of iPadOS, macOS etc)

ebaauw:
Interesting.

Do you know if an accessory with this service can be bridged?

Does an accessory with this service show in other HomeKit apps (than Apple's Home app)?  AirPlay2 devices only show in Home.

> MediaStates kind of work in the Home App (when switching between playing and paused the loading indicator will stay though).

You probably need to set `CurrentMediaState` to the value of `TargetMediaState` to change the indicator.  That's how it works for `CurrentPosition` and `TargetPosition` in the `WindowCovering` service.

> Volume and Mute Characteristics seem to not have any effect.

Do you mean: they're not visible in Home?  Or are they visible but not working?  Did you try the hardware `+` and `-` buttons in the _Remote_ widget?

Supereg:
> Do you know if an accessory with this service can be bridged?

Did not test this yet.

> Does an accessory with this service show in other HomeKit apps (than Apple's Home app)? 

The Eve App didn't show the Speaker (but the Eve App does also not show Televisions)

> You probably need to set CurrentMediaState to the value of TargetMediaState to change the indicator. That's how it works for CurrentPosition and TargetPosition in the WindowCovering service.

I know that already from Locks and Windows. If you have a lock at the example accessory I did exactly that.

> Do you mean: they're not visible in Home? Or are they visible but not working? Did you try the hardware + and - buttons in the Remote widget?

HAP does query the current state of those. However there are no controls shown to alter these values. The media control widget seems to have something to do with AirPlay 2 and nothing with the Characteristics. Also I did not check if the device gets displayed in the Remote widget.


I also noticed, that the old speaker device (which is used to indicate 2way audio for cameras) gained new optional characteristics: VolumeControlType and VolumeSelector. Don't know what's that for, since the Volume Characteristic wasn't event utilized by HomeKit.

ebaauw:
> (but the Eve App does also not show Televisions)

As far as I can tell that's a limitation of HomeKit, not of Eve.  Other HomeKit apps simply don't get to see the accessories.

> > You probably need to set CurrentMediaState to the value of TargetMediaState to change the indicator.
> If you have a look at the example accessory I did exactly that.

Indeed.  It might be a timing issue, however.  Did you try the `service.setCharacteristic(Characteristic.CurrentMediaState, targetMediaState);` from within a `setImmediate()` or `setTimeout()`?

> I also noticed, that the old speaker device (which is used to indicate 2way audio for cameras) gained new optional characteristics: VolumeControlType and VolumeSelector.

These are used in the TV Speaker service to drive the _Remote_ Widget.  I've implemented these in homebridge-zp, when exposing zone players as Television accessories (to control the input selection from Home).

Supereg:
> Indeed. It might be a timing issue, however. Did you try the service.setCharacteristic(Characteristic.CurrentMediaState, targetMediaState); from within a setImmediate() or setTimeout()?

Nope. Maybe I can check it sometime later. I would assume that this service is strongly paired with AirPlay 2 (nothing else would I expect from Apple honestly). And that it tries to send any other command via the AirPlay protocol and thus the loading indicator stays.

> These are used in the TV Speaker service to drive the Remote Widget. I've implemented these in homebridge-zp, when exposing zone players as Television accessories (to control the input selection from Home).

Yeah, already noticed that. HAP-NodeJS implements the exact same Speaker service two times (as Speaker and as TelevisionSpeaker 😅).

ebaauw:
> HAP-NodeJS implements the exact same Speaker service two times (as Speaker and as TelevisionSpeaker

Damn, you're right, the same UUID (113).  I never noticed that before.

Supereg:
closing in favor for https://github.com/KhaosT/HAP-NodeJS/pull/799

PR #799, opened 2020-03-31
Updated services and characteristics to reflect the latest specification of iOS 13.4

Supereg:
This PR updates some services to the most recent definition.

Newly added services:
* SmartSpeaker (originally https://github.com/KhaosT/HAP-NodeJS/pull/798)
* PowerManagement
* TransferTransportManagement

Edit: Additionally added the service AccessControl (is already in the spec like at least last year)

### DIFF
--- a/src/lib/gen/HomeKit.ts
+++ b/src/lib/gen/HomeKit.ts
@@ -4320,6 +4320,31 @@ export class SmokeSensor extends Service {
 Service.SmokeSensor = SmokeSensor;
 
 /**
+ * Service "Smart Speaker"
+ */
+
+export class SmartSpeaker extends Service {
+
+  static UUID: string = '00000228-0000-1000-8000-0026BB765291';
+
+  constructor(displayName: string, subtype: string) {
+    super(displayName, SmartSpeaker.UUID, subtype);
+
+    // Required Characteristics
+    this.addCharacteristic(Characteristic.CurrentMediaState);
+    this.addCharacteristic(Characteristic.TargetMediaState);
+
+    // Optional Characteristics
+    this.addOptionalCharacteristic(Characteristic.Name);
+    this.addOptionalCharacteristic(Characteristic.ConfiguredName);
+    this.addOptionalCharacteristic(Characteristic.Volume);
+    this.addOptionalCharacteristic(Characteristic.Mute);
+  }
+}
+
+Service.SmartSpeaker = SmartSpeaker;
+
+/**
  * Service "Speaker"
  */
 
--- a/src/lib/gen/HomeKit.ts
+++ b/src/lib/gen/HomeKit.ts
@@ -4,6 +4,29 @@ import { Characteristic, Formats, Perms, Units } from '../Characteristic';
 import { Service } from '../Service';
 
 /**
+ * Characteristic "Access Control Level"
+ */
+
+export class AccessControlLevel extends Characteristic {
+
+  static readonly UUID: string = '000000E5-0000-1000-8000-0026BB765291';
+
+  constructor() {
+    super('Access Control Level', AccessControlLevel.UUID);
+    this.setProps({
+      format: Formats.UINT16,
+      perms: [Perms.NOTIFY, Perms.PAIRED_READ, Perms.PAIRED_WRITE],
+      maxValue: 2,
+      minValue: 0,
+      minStep: 1,
+    });
+    this.value = this.getDefaultValue();
+  }
+}
+
+Characteristic.AccessControlLevel = AccessControlLevel;
+
+/**
  * Characteristic "Accessory Flags"
  */
 
@@ -1465,7 +1488,7 @@ export class NightVision extends Characteristic {
     super('Night Vision', NightVision.UUID);
     this.setProps({
       format: Formats.BOOL,
-      perms: [Perms.READ, Perms.WRITE, Perms.NOTIFY]
+      perms: [Perms.READ, Perms.WRITE, Perms.NOTIFY, Perms.TIMED_WRITE]
     });
     this.value = this.getDefaultValue();
   }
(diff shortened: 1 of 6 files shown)


Supereg:
The AccessControl service is used to configure the "Allow Speaker & TV Access" for I think TV and SmartSpeaker services. The AccessControlLevel represents: 0: "Anyone on the same Network", 1: "Only people Sharing this Home", 2: "unknown".
The PasswordSetting characteristic is used to receive the currently set password and has the following structure.
Type: 0x01 | Length: N | UTF8 password string (only present if required)
Type: 0x02 | Length: 1 | boolean if password is required

Should the AccessControl Service only be accessible by Admin controllers? Would make sense.

ebaauw:
> Should the AccessControl Service only be accessible by Admin controllers? Would make sense.

Did we discover yet how to the HAP server can find out whether a home administrator or regular home user is connected?  Is this exposed by HAP-NodeJS and/or Homebridge?

Supereg:
Yes, this was actually my first PR for this project 
https://github.com/KhaosT/HAP-NodeJS/pull/702

Supereg:
And there’s already an API to add those restrictions to your Characteristic definition https://github.com/KhaosT/HAP-NodeJS/blob/9655550410bdf6656c2c7e6ae8615f822ab042e0/src/lib/Characteristic.ts#L358

ebaauw:
Thanks, I'll have a look at it.

I always figured administrator-only access was defined at service level, indicated by the presence of the `AdministoratorOnlyAccess` charactertistic.  See https://github.com/homebridge/homebridge/issues/1389

Supereg:
Didn’t quite understand the use of this characteristics. I know no device which uses this characteristic.

For Apple TV Remotes characteristics it was specified that certain writes/reads can only be done by admins (restriction at a per characteristic and per perms [notify, read, write] level). And the mentioned AdministratorOnlyAccess is not used for Remotes.

ebaauw:
I'm sorry @Supereg, I don't understand how I am supposed to use `accessRestrictedToAdmins` in custom characteristics.  `Access` isn't exported, neither is (a getter or setter for) `accessRestrictedToAdmins`.  Why hasn't it been included in the characteristic properties?

Supereg:
Looking at the Code Access should be exported though?
https://github.com/KhaosT/HAP-NodeJS/blob/9655550410bdf6656c2c7e6ae8615f822ab042e0/src/lib/Characteristic.ts#L72-L76

To set restrictions you would just write to the `accessRestrictedToAdmins` property or push new ones. Here is the Apple TV Remote example:
https://github.com/KhaosT/HAP-NodeJS/blob/9655550410bdf6656c2c7e6ae8615f822ab042e0/src/lib/gen/HomeKit-Remote.ts#L27-L47

Supereg:
Note, that homebridge only supports this in the current beta version. As the last stable homebridge version ships with a hap-nodejs version, which doesn't support that yet.

ebaauw:
> Looking at the Code, `Access` should be exported though?

Indeed, but it's not exposed as a `static` property of `Characteristic` like `Formats`, `Units`, and `Perms`.
https://github.com/KhaosT/HAP-NodeJS/blob/9655550410bdf6656c2c7e6ae8615f822ab042e0/src/lib/Characteristic.ts#L144

> To set restrictions you would just write to the `accessRestrictedToAdmins` property

OK.  Feels kinda dirty.  Is that even allowed by TypeScript?

> Note, that homebridge only supports this in the current beta version. As the last stable homebridge version ships with a hap-nodejs version, which doesn't support that yet.

Thanks for the heads-up.  I am running the latest beta (testing my plugins for compatibility).

Supereg:
> Indeed, but it's not exposed as a static property of Characteristic like Formats, Units, and Perms, so it's not passed to the plugin by Homebridge.

AFAIK this is only to provide backwards compatibility, in order to access it like `Characteristic.Formats`. The Access enum (and also all the other enums) are exported at the same level as the Characteristic class. There is not own namespace for that.

> OK. Feels kinda dirty. Is that even allowed by TypeScript?

Well yes, it's a normal 'class attribute' without a private access modifier. Somewhat the idea was, that CharacteristicProps are all HAP defined attributes (which are also returned on a discover request), while the admin access isn't really one. But we can also move that into the Properties 🤷🏽‍♂️

Here is an example how you would declare a custom Characteristic with the correct imports at the top.
```js
import {Access, Characteristic, Formats, Perms, Service} from "hap-nodejs";

class CustomCharacteristic extends Characteristic {

    static readonly UUID: string = '00000000-0000-0000-0000-00000000';

    constructor() {
        super('Custom Characteristic', CustomCharacteristic.UUID);
        this.setProps({
            format: Formats.UINT8,
            perms: [Perms.PAIRED_WRITE, Perms.PAIRED_READ]
        });
        this.value = this.getDefaultValue();
        this.accessRestrictedToAdmins = [Access.READ, Access.WRITE];
    }
    
}

const service = new Service.Lightbulb('', '');
const existingCharacteristic  = service.getCharacteristic(Characteristic.On)!;

existingCharacteristic.accessRestrictedToAdmins.push(Access.READ);
existingCharacteristic.accessRestrictedToAdmins = [Access.READ];
```

Supereg:
Additionally there is a new AUDIO_RECEIVER device category.

src/lib/gen/HomeKit.ts
```diff
@@ -4214,10 +4340,10 @@ export class Outlet extends Service {
 
     // Required Characteristics
     this.addCharacteristic(Characteristic.On);
-    this.addCharacteristic(Characteristic.OutletInUse);
```
KhaosT, line 4217: Let's keep this as required for backwards compatibility (older iOS versions)

KhaosT:
Just one minor thing :) Thanks!

Supereg:
Resolved. Additionally reverted my updated definition of the Speaker service. As it's already covered by the TelevisionSpeaker service, which is essentially the same.

KhaosT:
👍

PR #802, opened 2020-04-02
Introduce newly engineered Controller API and CameraController API

Supereg:
So, okay. Whats this for?
This PR introduces a new abstraction layer called `Controller`, which basically represents an arrangement of (multiple) services. Those services are generally seen as one bigger device (like currently supported: Cameras/Doorbells, Apple TV Remotes). Controllers are also intended to provide some simple to use API (for the end user) for more complex services and characteristics (In theory I got plans to add controllers for TVs, Routers [if we get the protocol sometime] and maybe for locks with a LockManagement service). This new abstraction layer unifies handling and declares certain specifications for devices like this.
This interface builds upon those newly introduced serialize and deserialize functions. It is build from ground up to support serialization and deserialization of services and characteristics from persistent storage. This is one of the features needed for homebridge to fully support bridged cameras and Apple TV remotes. Additionally so called `SerializableControllers` can be used to gain persistent storage of various configuration or data (for example Apple TV target configurations; or for the future for Secure Video). This is automatically enabled for supported Controllers using the new `ControllerStorage` class.
A second design decision is, that any Controller is somewhat decoupled from an Accessory, meaning a Controller does not need a reference to the Accessory to which it is added. This is pretty nice when used with homebridge. This means though, that this PR currently heavily depends on https://github.com/KhaosT/HAP-NodeJS/pull/786.

So that's the general/abstract stuff.

----

One implementation of a Controller is the newly introduced `CameraController` (and `DoorbellController`). It provides a refined way of creating IPCameras, which improved config types (more optional and default stuff) and Secure Video in mind, to provide a smooth transition once this get implemented. It automatically creates all necessary/configured `CameraRTPStreamManagement` services. The `StreamController` was renamed to `RTPStreamManagement` (for obvious naming collisions, backwards compatibility is ensure though) and has undergone some heavy code cleanup and refinements (with some little fixes and improvements to the tlv structure).
The old style `CameraSource` got replaced by the new `CameraStreamingDelegate`, which defines a much more minimal interface, with improved typing and a new callback interface to better signal errors to the requesting HomeKit device.

With all this said, it is important to note that **backwards compatibility is fully provided**. A old style CameraSource gets wrapped into an Adapter, a CameraController is created and it basically just works. Meaning even those old CameraSource will gain the new service serialization/deserialization feature. All by the CameraSource created StreamControllers and Camera related services are ignored and get replaced by the ones of the CameraController.
The only thing you won't get are the new features provided through `CameraStreamingDelegate`.

The old `Camera.ts` example got replaced by a new `Camera_accessory.ts` using the new API.

----

Lastly this PR contains also some minor code cleanup and improvements at some other places.
Additionally some refactoring was made (renaming and relocating stuff), but without breaking anything.

All those changes are the result of weeks thinking and prototyping (many ideas where tried and scrapped again). One could call the result a bit over engineered lol. But I think the result was worth it. I'm happy to hear your feedback and maybe ways to improve it further.

All changes were on the one hand tested with one of my Doorbell implementations to ensure no regressions were made. Additionally I checked some of the bigger camera plugins (like the ring plugin) to ensure they still are compatible API wise.

Happy to get your review
Andi


Supereg:
Latest force-push fixes a misconception I had made in ControllerStorage. Now works for bridged accessories.

Any feedback, improvements or concerns for this PR are welcome.

KhaosT:
Thanks ^^ I'll try to take a look over the weekend... Sorry for the slow PR reviews 😅

KhaosT:
And for backwards compatibility reasons, I'm not sure we want to just delete `StreamController` and break everything that depend on this out in the wild right now... Preferably we may keep both ways in the library for now, and encourage people to switch to adopt the new way?

Supereg:
StreamController isn't actually deleted. Git seems to be just very bad to track moved and renamed files. StreamController basically underwent some adjustments and improvements and is now called RTPStreamManagement (to better reflect what it does and to get that 'Controller' out of the name as this would collide with the new concept of Controllers).
However there is still a class StreamController which basically is an empty class which inherits from RTPStreamManagement to still provide said backwards compatibility.
The file got move into the camera folder but is still normally exported in the index.ts, so this won't change anything. The only "breaking" change I done to it was to rename some of the TLV definitions present since the Typescript refactoring.

Supereg:
As said in the text, this PR is FULLY backwards compatible. Only thing is services are ignored (everything which is a camera service) and the streamController array just gets replaced with the correctly initialized ones.

Supereg:
Changes are released on a beta release for now and incorporated into the homebridge beta soon.

PR #806, opened 2020-04-06
Adding getServiceById to retrieve Service by UUID and subtype

Supereg:
This basically move homebridges `getServiceByUUIDAndSubType` into hap-nodejs.

https://github.com/homebridge/homebridge/blob/e67630a3978aa126786711f8ec06610db517ca74/lib/platformAccessory.js#L113-L122


KhaosT:
🙏

PR #807, opened 2020-04-08
Adding ability to replace characteristics of a service with the characteristics of another service (of the same service type)

Supereg:
* Adding ability to replace characteristics of a service with the characteristics of another service (of the same service type)
* Fixed some typings


Supereg:
Open again. Though I messed something up, but everything works. Is now fully tested.

PR #811, opened 2020-04-11
Add helpers for async characteristic get/set handlers

samuelthomas2774:
This adds two methods for using async functions as characteristic get/set handlers. This also works for get/set handlers that don't make any asynchronous calls so they can just return/throw instead of using a callback.

#### `characteristic.getHandler(...)`, `characteristic.setHandler(...)`

```ts
characteristic.getHandler(async () => {
    // Get the value asynchronously...
});
characteristic.setHandler(async value => {
    // Set the value asynchronously...
});
```

#### Extending the Characteristic class/a predefined Characteristic

```ts
class FanOnCharacteristic extends Characteristic.On {
    constructor(public readonly fan: Fan) {
        super();
    }

    async [Characteristic.GetValueHandler]() {
        // Get the value asynchronously...
    }

    async [Characteristic.SetValueHandler]() {
        // Set the value asynchronously...
    }
}
```

### DIFF
--- a/src/accessories/Fan_accessory.ts
+++ b/src/accessories/Fan_accessory.ts
@@ -6,97 +6,122 @@ import {
   Characteristic,
   CharacteristicEventTypes, CharacteristicSetCallback,
   CharacteristicValue,
+  Formats,
+  Perms,
   NodeCallback,
   Service,
   uuid,
-  VoidCallback
+  VoidCallback,
 } from '..';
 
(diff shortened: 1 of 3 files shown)


Supereg:
Thanks for the PR.
This PR was obsoleted by https://github.com/homebridge/HAP-NodeJS/pull/849, as it provides a bit better naming for setters and getters and integrates better with the latest beta changes.

PR #819, opened 2020-05-05
Add missing required characteristic of InputSource

jonavila:
The linked InputSource services of the Television service were missing the required characteristic `Characteristic.CurrentVisibilityState`


Supereg:
The CurrentVisibilityState is a required characteristic of InputSource and thus will be automatically added to the InputSource characteristic with a default value of `0` which in fact represents `SHOWN`.
So there is no need actually to add it here.

But we could add it anyways to show it's existence in the example 🤔

jonavila:
@supereg if I don’t add it explicitly it doesn’t show up for me on the Home app.

Supereg:
Did not have a problem running the example here. Could you verify that it was indeed an issue?

jonavila:
@Supereg Sorry for the confusion. I thought this was an issue because the home app doesn't seem to update accordingly with the services and characteristics unless you completely close it out.

PR #836, opened 2020-07-06
Update index.ts

n0rt0nthec4t:
RemoteController was duplicated in this file. Think DoorbellController should be here?

### DIFF
--- a/src/lib/controller/index.ts
+++ b/src/lib/controller/index.ts
@@ -1,4 +1,4 @@
 export * from './Controller';
 export * from './RemoteController';
 export * from './CameraController';
-export * from './RemoteController';
+export * from './DoorbellController';


Supereg:
Thanks for noticing.

n0rt0nthec4t:
No worries.. Since you developed the new controller classes, wondering how you can override the default name of the services created.. ie:, both camera and doorbell service have "" as the name, however I'd like to set this at creation time. Setting the Characteristic.Name after doesn't work..

Supereg:
sorry, completely missed your response. AFAIK know the name is taken from the AccessoryInformation service (aka the accessory display name).

As the camera service (RTP Stream management) doesn't have an Name characteristic (according to apple). Only Doorbell service may have an optional Name characteristic. For that you seem to be correct, that the current code doesn't allow setting it 🤔

PR #849, opened 2020-10-11
Add support for promise based Characteristic setters and getters

oznu:
I didn't see the other PR until after I went to make this PR...

Usage (async is optional):

```ts
fan
  .getService(Service.Fan)!
  .addCharacteristic(Characteristic.RotationSpeed)
  .onGet(async () => {
    return FAKE_FAN.rSpeed;
  })
  .onSet(async (value) => {
    FAKE_FAN.setSpeed(value);
  });
```

The user can also throw a specific HAP error:

```ts
fan
  .getService(Service.Fan)!
  .addCharacteristic(Characteristic.RotationSpeed)
  .onGet(async () => {
    throw new HapStatusError(HAPStatus.OPERATION_TIMED_OUT);
  });
```

If any other error types are thrown while executing the handler, it will fallback to `HAPStatus.SERVICE_COMMUNICATION_FAILURE`.

`HapStatusError` can also be used in the current style listeners:

```ts
fan
  .getService(Service.Fan)!
  .addCharacteristic(Characteristic.RotationSpeed)
  .on('get', (callback) => {
    callback(new HapStatusError(HAPStatus.OPERATION_TIMED_OUT));
  });
```

Characteristic that require a  `WRITE_RESPONSE` can return the value in their `onSet` handler. Returned values for characteristics that do not implement `WRITE_RESPONSE` (99% of them) are ignored.

```ts
fan
  .getService(Service.Fan)!
  .addCharacteristic(Characteristic.RotationSpeed)
  .onSet(async (value) => {
    return 99;
  });
```

If the user attempts to setup both an `onSet` / `onGet` handler and the current style `.on('get' / 'set')` listeners, the `onSet` / `onGet` will take priority and a warning will be displayed in the log.

### DIFF
--- a/src/accessories/Fan_accessory.ts
+++ b/src/accessories/Fan_accessory.ts
@@ -4,9 +4,7 @@ import {
   AccessoryEventTypes,
   Categories,
   Characteristic,
-  CharacteristicEventTypes, CharacteristicSetCallback,
   CharacteristicValue,
-  NodeCallback,
   Service,
   uuid,
   VoidCallback
@@ -62,9 +60,8 @@ fan.on(AccessoryEventTypes.IDENTIFY, (paired: boolean, callback: VoidCallback) =
 fan
   .addService(Service.Fan, "Fan") // services exposed to the user should have "names" like "Fake Light" for us
   .getCharacteristic(Characteristic.On)!
-  .on(CharacteristicEventTypes.SET, (value: CharacteristicValue, callback: CharacteristicSetCallback) => {
+  .onSet((value) => {
     FAKE_FAN.setPowerOn(value);
-    callback(); // Our fake Fan is synchronous - this value has been successfully set
   });
 
 // We want to intercept requests for our current power state so we can query the hardware itself instead of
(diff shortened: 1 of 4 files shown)


Supereg (approved):

Kosta-Github:
Could this also be merged into the main branch in addition to the beta branch?

Supereg:
This change is part of the 0.9.0 release and won't be back ported to the 0.8.x line. It is a API change and thus needs a minor version increment.

MiguelRipoll23:
> 'HapStatusError' cannot be used as a value because it was exported using 'export type'.

If thrown in an exception is used as a value but it's exported as a type, maybe I'm doing something wrong? This API is cleaner and looks fancier by the way so thank you for these promise-based getters/setters.

Supereg:
@MiguelRipoll23 when using it in homebridge you need to access it via the `.hap.HAPStatusError` property of the homebridge api object.

MiguelRipoll23:
@Supereg thanks, that was it!

PR #854, opened 2020-11-10
Catch validation errors in set update

oznu:
If an existing homebridge plugin were to provide an invalid values using the `.setValue` or `.updateValue` methods the bridge would crash without reporting which plugin or characteristic caused the problem.

After this change, it should handle invalid values the same way we do if one is provided by the on set/get handlers.

### DIFF
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -1031,7 +1031,16 @@ export class Characteristic extends EventEmitter {
       callback = undefined;
     }
 
-    value = this.validateUserInput(value)!;
+    try {
+      value = this.validateUserInput(value)!;
+    } catch (e) {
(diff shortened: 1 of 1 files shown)


oznu:
These changes should not have caused the build to fail - it's not even getting to the testing step.

The tests are passing locally.

Supereg:
> These changes should not have caused the build to fail - it's not even getting to the testing step.

some npm command not found 🤔

oznu:
Yeah I don't know whats going on. I re-ran one of the previous builds and it works fine, so it does not seem as if Github Actions has changed their environments.

Supereg:
closed via https://github.com/homebridge/HAP-NodeJS/commit/f0e8708713025067294854170ce9b67a484b0cb3

PR #855, opened 2020-11-12
Replace term 'Ambient lightning' by 'Adaptive Lightning'

Zer0x00:
Apple's terminology for the adapting light mode is 'Adaptive Lightning', seems like you (@Supereg) had a little mistake there.

I've replaced every term I could find in the files and in the filenames.

The procedure was:
1. Search in all files and replace 'ambient' by 'adaptive'
2. Search in all files and replace 'Ambient' by 'Adaptive'
3. Search all filenames and replace 'ambient' by 'adaptive'
4. Search all filenames and replace 'Ambient' by 'Adaptive'
(all case-sensitive)


llamafilm:
Shouldn't this say Lighting instead of Lightning?

Zer0x00:
> Shouldn't this say Lighting instead of Lightning?

Thx, should be fixed.

Supereg:
The characteristic `CurrentAmbientLightLevel` was mistakenly renamed to `CurrentAdaptiveLightLevel` by this PR.
Just quickly implemented the change myself with https://github.com/homebridge/HAP-NodeJS/commit/0a8fee6536a15e480a3b624c6c01ee69f9a65a83.

Big thank you for catching that 😅

PR #866, opened 2021-01-06
Fix incorrect bounds checks of HAPStatus errors

rutzfutz:
The upper/lower bounds of the checks were reversed due to negative number confusion. I also refactored the duplicate checks into a single IsKnownHAPStatusError() function.

### DIFF
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -237,7 +237,7 @@ import {
   WiFiConfigurationControl,
   WiFiSatelliteStatus,
 } from "./definitions";
-import { HAPStatus } from "./HAPServer";
+import { HAPStatus, IsKnownHAPStatusError } from "./HAPServer";
 import { IdentifierCache } from './model/IdentifierCache';
 import { Service } from "./Service";
 import { clone } from "./util/clone";
@@ -2146,7 +2146,7 @@ function extractHAPStatusFromError(error: Error) {
   if (numberPattern.test(error.message)) {
     const value = parseInt(error.message);
 
-    if (value >= HAPStatus.INSUFFICIENT_PRIVILEGES && value <= HAPStatus.NOT_ALLOWED_IN_CURRENT_STATE) {
+	if (IsKnownHAPStatusError(value)) {
       errorValue = value;
     }
   }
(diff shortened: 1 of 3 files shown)


rutzfutz:
Fixes #867.

Supereg:
Thanks for fixing that 👍 I like how you moved the bounds check to the enum definition (although the export will kind of make it part of the public API, but that's okay I think). This should allow for easier updating.
Nonetheless I will add a regression test for this myself.

rutzfutz:
> EDIT: I think, we can even write a test case which catches the case, where a new error code is added but the bounds checking function is not adjusted.

I tried testing for this by iterating through all values of the `HAPStatus` enum and expecing my new `IsKnownHAPStatusError` function to return true for all of them, but couldn't iterate through the values of a `const enum`. Apparently, the TypeScript compiler completely eliminates a `const enum`, replacing references to it with the actual values, which makes it impossible to iterate through all enum values. But maybe you are more successful. 😅

Supereg:
Yes, we switched to `const enums` once we migrated homebridge to typescript. This allows plugin devs to import those as types (and they do not need to access homebridge api for that, e.g `api.hap.HAPStatus`). The Typescript compiler will always replace const enums with the value in the code.
Thus the default behavior for const enums is also, that the resulting javascript code does not contain any enum definition anymore. Thats the reason why the Typescript compiler warns you about not being able to use a `const enum` as a value, as it doesn't exists.

However as we need to retain the actual `const enum` as a value for backwards compatibility, we turned on `"preserveConstEnums": true` compiler option, which does not remove the `const enum` from the javascript code. Thus we can also forcibly access `const enums` as values. Nonetheless the Typescript compiler will mark those as errors. Only way out is `@ts-ignore` or `@ts-expect-error`.

rutzfutz:
> However as we need to retain the actual `const enum` as a value for backwards compatibility, we turned on `"preserveConstEnums": true` compiler option, which does not remove the `const enum` from the javascript code. Thus we can also forcibly access `const enums` as values. Nonetheless the Typescript compiler will mark those as errors. Only way out is `@ts-ignore` or `@ts-expect-error`.

Ha, gotcha! I didn't know about this compiler option. Just saw your regression test. Great work!

Issue #872, opened 2021-01-27
Cannot add a bridged Accessory with the same UUID as another bridged Accessory: undefined


macinspak:
**Describe the bug**
Difficult to reproduce, I have it currently. I noticed it first in node-red-contrib-homekit-bridged, but after debugging for a while, it seems that although NRCHKB is only calling AccessoryUtils.getOrCreate once, the loop **inside HAP-NodeJS** to add it internally to the accessories in the bridge is looping twice and seems to have an index of "NaN" already in the fresh bridge with an empty array of accessories, which is causing it to "match" a non-existent UUID in the bridge and complain about an undefined UUID (when the UUID is defined in the accessory).

This seemed to "appear" in an installation I had been running successfully for a long time, then (all of a sudden, no changes, yeah- I know) and I was getting "no response" on all devices in homekit, so I removed the bridge from home.app but now cannot re-add the bridge, it doesn't appear in the home.app at all.

a lot of debug leading to this issue being raised is documented in [NRCHKB Issue 359](https://github.com/NRCHKB/node-red-contrib-homekit-bridged/issues/359)

**To Reproduce**
Seems to be only on my server, someone else running the same flow (as per [test flow](https://github.com/NRCHKB/node-red-contrib-homekit-bridged/issues/359#issuecomment-758411019)) seems to not get the issue.

**Desktop (please complete the following information):**
 - OS: Rpi
 - Node version: v14.15.4
 - NPM version: 6.14.10
 - NodeRed version: v1.2.7
 - NRCHKB Plugin Version: 1.2.0
 - NRCHKB uses HAP-NodeJS Version: 0.7.8

**Additional context**
running debug

```
DEBUG=NRCHKB*,HAP-NodeJS*,Accessory,HAPServer,EventedHTTPServer node-red
```

I get a lot of these first in the debug output:

```
  NRCHKB:HAPServiceNode Starting Parent Service Pool 1 +6ms
  NRCHKB:AccessoryUtils Looking for accessory with service subtype '172f1edd-389b-4c45-96d4-47e37a130118'... +0ms
  NRCHKB:AccessoryUtils ... didn't find it. Adding new accessory with name 'Pool 1' and UUID '1316cdc7-e142-4a55-9a93-fea8fce35aa5' +0ms
  NRCHKB:HAPServiceNode Starting Parent Service Pool 2 +6ms
[...]

PR #879, opened 2021-02-14
Refactor for...in to for...of

otaku:
## Original Problem
for...in will iterate inherited enumerable properties.  Other nodes added properties to Array and were being iterated as well causing strange errors.

## Solution
The following refactors were applied:

- for...in was changed to for...of.  In places where Typescript complains, a guard was added.
- added a radix of 10 to all parseInt calls
- removed unnecessary escape characters in regular expression strings
- removed unnecessary semicolons in class function declaration
- changed Typescript namespace to ES2015 class

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -437,8 +437,7 @@ export class Accessory extends EventEmitter {
         : serviceParam;
 
     // check for UUID+subtype conflict
-    for (let index in this.services) {
-      const existing = this.services[index];
+    for (let existing of this.services) {
       if (existing.UUID === service.UUID) {
         // OK we have two Services with the same UUID. Check that each defines a `subtype` property and that each is unique.
         if (!service.subtype)
@@ -521,9 +520,7 @@ export class Accessory extends EventEmitter {
   }
 
   public getServiceById<T extends WithUUID<typeof Service>>(uuid: string | T, subType: string): Service | undefined {
-    for (const index in this.services) {
-      const service = this.services[index];
-
+    for (const service of this.services) {
       if (typeof uuid === "string" && (service.displayName === uuid || service.name === uuid) && service.subtype === subType) {
         return service;
       } else if (typeof uuid === "function" && ((service instanceof uuid) || (uuid.UUID === service.UUID)) && service.subtype === subType) {
@@ -591,8 +588,7 @@ export class Accessory extends EventEmitter {
   }
 
   public addBridgedAccessories(accessories: Accessory[]): void {
-    for (let index in accessories) {
-      const accessory = accessories[index];
+    for (const accessory of accessories) {
       this.addBridgedAccessory(accessory, true);
     }
 
@@ -603,19 +599,15 @@ export class Accessory extends EventEmitter {
     if (accessory._isBridge)
       throw new Error("Cannot Bridge another Bridge!");
 
-    let foundMatchAccessory = false;
     // check for UUID conflict
-    for (let index in this.bridgedAccessories) {
-      const existing = this.bridgedAccessories[index];
-      if (existing.UUID === accessory.UUID) {
-        foundMatchAccessory = true;
-        this.bridgedAccessories.splice(Number.parseInt(index), 1);
-        break;
-      }
-    }
+    const foundMatchAccessory = this.bridgedAccessories.findIndex((existing) => {
+      return existing.UUID === accessory.UUID
+    });
 
-    if (!foundMatchAccessory)
+    if (foundMatchAccessory === -1)
       throw new Error("Cannot find the bridged Accessory to remove.");
+
+    this.bridgedAccessories.splice(foundMatchAccessory, 1);
 
     accessory.removeAllListeners();
 
@@ -625,8 +617,7 @@ export class Accessory extends EventEmitter {
   }
 
   public removeBridgedAccessories(accessories: Accessory[]): void {
-    for (let index in accessories) {
-      const accessory = accessories[index];
+    for (const accessory of accessories) {
       this.removeBridgedAccessory(accessory, true);
     }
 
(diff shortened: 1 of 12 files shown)


src/lib/controller/RemoteController.ts
```diff
@@ -1066,11 +1064,13 @@ export class RemoteController extends EventEmitter implements SerializableContro
 
     private handleDataStreamConnectionClosed(connection: DataStreamConnection): void {
         for (const targetIdentifier in this.dataStreamConnections) {
```
Supereg, line 1066: Can this also be transformed to a `for ... of`? 
Using `for (const [targetIdentifier, connection0] of Object.entries(this.dataStreamConnections))` so you are able to do the `delete`?

Supereg (commented):
Appart from the one comment PR looks good. Thanks for the contribution 👍 🎉 
Maybe you can quickly elaborate on that specific decision, otherwise happy to merge.

Supereg:
Did not have a chance yet to read through #872. Can this issu? be closed once this PR is merged?

src/lib/Accessory.ts
```diff
@@ -437,8 +437,7 @@ export class Accessory extends EventEmitter {
         : serviceParam;
 
     // check for UUID+subtype conflict
-    for (let index in this.services) {
-      const existing = this.services[index];
+    for (let existing of this.services) {
```
Shaquu, line 440: Can it be const?

Shaquu (commented):
One small comment from my side also.
I noticed one TODO, is it possible to resolve in this PR?

Shaquu:
@Supereg It should resolve issue from what first reporter said.

NRCHKB/node-red-contrib-homekit-bridged#359

src/lib/controller/RemoteController.ts
otaku: The issue is with Typescript actually.  Object.keys and Object.entries both return key as string.

As an alternative, I converted Record into a Map.  I was considering applying the same change to other Record variables as well if this is acceptable.

src/lib/controller/RemoteController.ts
```diff
@@ -58,7 +58,7 @@ const enum SupportedButtonConfigurationTypes {
     BUTTON_TYPE = 0x02
 }
 
-export const enum ButtonType {
```
otaku, line 61: I removed the `const` here to allow the removal of a `@ts-ignore`

src/lib/controller/RemoteController.ts
```diff
@@ -58,7 +58,7 @@ const enum SupportedButtonConfigurationTypes {
...
+export enum ButtonType {
```
Supereg, line 61: I would like to keep this a `const enum` If possible. This has some implications on homebridge plugins, as using const enums, they can easily just use type imports. For JavaScript compatibility we also have turned on the „preserve const enums“ transpiler option.
I know this makes usage of this enum a bit painful and requires some `@ts-ignore` or `@ts-expect-error` in some places.

src/lib/controller/RemoteController.ts
```diff
@@ -1063,14 +1071,11 @@ export class RemoteController extends EventEmitter implements SerializableContro
...
     private handleDataStreamConnectionClosed(connection: DataStreamConnection): void {
-        for (const targetIdentifier in this.dataStreamConnections) {
-            if (Object.prototype.hasOwnProperty.call(this.dataStreamConnections, targetIdentifier)) {
-                const connection0 = this.dataStreamConnections[targetIdentifier];
-                if (connection === connection0) {
-                    debug("HDS connection disconnected for targetIdentifier %s", targetIdentifier);
-                    delete this.dataStreamConnections[targetIdentifier];
-                    break;
-                }
+        for (const [ targetIdentifier, connection0 ] of this.dataStreamConnections) {
```
Supereg, line 1074: Like the migration to `Map`, makes this a lot simpler 👍

src/lib/controller/RemoteController.ts
Supereg: See https://github.com/homebridge/HAP-NodeJS/pull/879#discussion_r575848556

otaku:
Would you like more refactoring applied to this PR or opened in a new PR?  I'd be targeting Record to Map specifically.

src/lib/controller/RemoteController.ts
Supereg: Ah yeah, makes sense.

If you like, you can migrate that in other places as well, though maybe doing this in a separate PR? So we can merge this, as it seemingly solves some important issues?

Is there any disadvantage to regular objects/Records? Like performance wise? I‘m not really educated on Map vs Record.

Supereg (dismissed):

otaku:
I noticed I serialized a Map, which JSON.stringify will fail to do.

src/lib/controller/RemoteController.ts
```diff
@@ -346,7 +346,7 @@ interface RemoteControllerServiceMap extends ControllerServiceMap {
 
 interface SerializedControllerState {
     activeIdentifier: number,
-    targetConfigurations: Map<number, TargetConfiguration>;
+    targetConfigurations: [number, TargetConfiguration][];
```
Supereg, line 349: `targetConfigurations` was a Record previously right? Is this still compatible with previous created storage entries?

otaku:
Thank you, I restored the type and serialize / deserialize from Record to Map instead.  Let me know if I should just rip out Map instead? Though it may still be better as it's less prone to enumerating inherited properties.

Supereg (approved):
Even if serialization and deserialization isn't as clean, I think having Map is nonetheless an improvement 👍 
Thanks for the contribution.

Supereg:
can we merge the PR or is something still missing?

otaku:
Nothing else is missing. This PR is complete. Thank you!

PR #881, opened 2021-02-15
Incorrect rounding breaking temperature values

NorthernMan54:
Saw this with my temperature sensors that are used outdoor.

```
HAP-NodeJS:Accessory [Bart-Dev-1 C734] Got Characteristic "Current Temperature" value: "-2.299999999999997"
```
And after delving into the code found this issue with javascript math and negative or decimal values. This issue is only triggered when the minValue for a characteristic is a negative or decimal number.

Added 2 test cases to cover the identified scenarios

The fix is a bit of kludge versus adding in a proper math library.

### DIFF
--- a/src/lib/util/request-util.spec.ts
+++ b/src/lib/util/request-util.spec.ts
@@ -49,4 +49,22 @@ describe("request-util", () => {
     });
     expect(formatOutgoingCharacteristicValue(36.135795, props)).toBe(36.1);
   });
+
+  it("should handle negative minumum values", () => {
+    const props = createProps(Formats.INT, {
+      minStep: 0.1,
+      minValue: -100,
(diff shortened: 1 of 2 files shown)


Supereg (approved):
I assume we have to revisit this part of the code in the future. We will never achieve the wanted accuracy (in regards to the step value) with using floats.
Maybe there is some neat solution out there to do this with a string base approach,
as we really only care about a proper string representation, rounded to the closes step value (starting at the minValue).

But for now, I think this is fair enough for an intermediate fix, yielding good results for the most common cases. Thanks for the contribution 👍

NorthernMan54:
Tks

My bad on the first PR I wasn’t aware of the automated testing going on.   Looks really

PR #884, opened 2021-02-21
Update validateUserInput to fix invalid values in more scenarios

oznu:
Throw more warnings instead of errors, if invalid value is supplied, revert back to last known good value, or a default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "hap-nodejs",
-  "version": "0.9.2-beta.6",
+  "version": "0.9.2-beta.7",
+  "betaVersion": "0.9.2",
   "description": "HAP-NodeJS is a Node.js implementation of HomeKit Accessory Server.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.9.1",
+  "version": "0.9.2-beta.2",
   "description": "HAP-NodeJS is a Node.js implementation of HomeKit Accessory Server.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.9.2-beta.2",
+  "version": "0.9.2-beta.4",
   "description": "HAP-NodeJS is a Node.js implementation of HomeKit Accessory Server.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.9.2-beta.4",
+  "version": "0.9.2-beta.6",
   "description": "HAP-NodeJS is a Node.js implementation of HomeKit Accessory Server.",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -1283,7 +1283,7 @@ export class Characteristic extends EventEmitter {
     try {
       value = this.validateUserInput(value)!;
     } catch (error) {
-      this.characteristicWarning(error.message + "", CharacteristicWarningType.ERROR_MESSAGE, error?.stack);
+      this.characteristicWarning(error?.message + "", CharacteristicWarningType.ERROR_MESSAGE, error?.stack);
       if (callback) {
         callback(error);
       }
@@ -1364,7 +1364,7 @@ export class Characteristic extends EventEmitter {
     try {
       value = this.validateUserInput(value);
     } catch (error) {
-      this.characteristicWarning(error.message + "", CharacteristicWarningType.ERROR_MESSAGE, error?.stack);
+      this.characteristicWarning(error?.message + "", CharacteristicWarningType.ERROR_MESSAGE, error?.stack);
       if (callback) {
         callback();
       }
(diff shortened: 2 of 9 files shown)


src/lib/definitions/generator-configuration.ts
```diff
@@ -75,6 +75,9 @@ export const CharacteristicOverriding: Map<string, (generated: GeneratedCharacte
   ["rotation.speed", generated => {
     generated.units = "percentage";
   }],
+  ["temperature.current", generated => {
+    generated.minValue = -273.15;
+  }],
```
oznu, line 80: Absolute zero.

image

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hap-nodejs",
-  "version": "0.9.1",
+  "version": "0.9.2-beta.6",
```
Supereg, line 3: I saw beta 6 was already released? where is the code for that? why is it not pushed to beta branch?
The CI for the beta branch currently works with having a `betaVersion` property defining what the base of the beta version is. In that case it would be `0.9.2`. This way it will always bull in the latest tag number.

```suggestion
  "version": "0.9.1",
  "betaVersion": "0.9.2",
```

src/lib/Characteristic.ts
```diff
@@ -1256,7 +1283,7 @@ export class Characteristic extends EventEmitter {
     try {
       value = this.validateUserInput(value)!;
     } catch (error) {
-      this.characteristicWarning(error.stack + "", CharacteristicWarningType.ERROR_MESSAGE);
+      this.characteristicWarning(error.message + "", CharacteristicWarningType.ERROR_MESSAGE);
```
Supereg, line 1286: Would it maybe make sense to extend characteristicWarning method to be able to overwrite the stack trace if we got one, like in this case?

src/lib/Characteristic.ts
```diff
@@ -1337,7 +1364,7 @@ export class Characteristic extends EventEmitter {
     try {
       value = this.validateUserInput(value);
     } catch (error) {
-      this.characteristicWarning(error.stack + "");
+      this.characteristicWarning(error.message + "");
```
Supereg, line 1367: Passing custom stack/error to characteristic warning, same as above.

src/lib/Characteristic.ts
```diff
@@ -1686,7 +1713,12 @@ export class Characteristic extends EventEmitter {
       case Formats.ARRAY:
         return [];
       default:
-        return this.props.minValue ?? 0;
+        switch(this.UUID) {
+          case Characteristic.CurrentTemperature.UUID:
+            return 0;
```
Supereg, line 1718: What is the rationale behind this special case?
Only to not deliver -273?

src/lib/Characteristic.ts
```diff
@@ -1686,7 +1713,12 @@ export class Characteristic extends EventEmitter {
...
+          default:
+            return this.props.minValue ?? 0;
```
Supereg, line 1720: Just a though while reading through that.
We might need to consider cases where `validValues` or `validValueRanges` is defined but not `minValue`.
I think you could use the `validValuesIterator()` and just pop the first element, to not create duplication logic.

src/lib/Characteristic.ts
```diff
@@ -1835,12 +1867,9 @@ export class Characteristic extends EventEmitter {
    * @param value - The value received from the API call
    */
   private validateUserInput(value?: Nullable<CharacteristicValue>): Nullable<CharacteristicValue> {
-    if (value === undefined) {
-      this.characteristicWarning(`characteristic was supplied illegal value: undefined! This might throw errors in the future!`);
-      return this.value; // don't change the value
-    } else if (value === null) {
```
Supereg, line 1841: Why would we want to allow undefined values? All cases I seen so far stem from programming errors which would have gone unnoticed otherwise. Am I missing something in the diff?

Edit: Saw that those get mapped to default values for certain values. As long as we still output a warning, I'm fine with that 👍

Supereg (commented):
Thanks for the pr. Also the added test cases are a added benefit to the project 👍 

Lefts some thoughts on some snippets.

src/lib/Characteristic.ts
oznu: calling `validValuesIterator()` here results in some sort of loop when running the tests.

oznu:
@Supereg should the Formats.FLOAT type have the same minimum and maximum values as Formats.INT?

src/lib/Characteristic.ts
oznu: In the interest of getting this out asap, I've just added some rudimentary checks here instead.

src/lib/Characteristic.ts
oznu: Updated the `characteristicWarning` method to take an optional stack trace. Added stack traces to everywhere that makes sense.

src/lib/Characteristic.ts
oznu: As above 😄

src/lib/Characteristic.ts
oznu: The first plugin i tried was getting the initial value from the default, but then set min props to `-50` so it triggered a warning that `-273` was less than the minimum of `-50`. Setting to zero just avoids any other potential issues that may arise from this change.

oznu: I've put a comment in the code explaining why.

src/lib/Characteristic.ts
oznu: Yep, the undefined handler is now right down the bottom in case it manages to get through all the other checks and still be undefined!

Supereg (commented):
You increased all the timeouts for characteristics reads/writes to 15 seconds? But the snapshot timeout is still at 10s?

What is the reason that we increase characteristics read and writes to 15 seconds?

Other than that the PR looks great to me 👍

oznu:
15 second total timeout was to reduce support requests for Homebridge 1.3.x. If people were happy with their "slow" setup prior we shouldn't force terminate the request on them to aggressively. The warning is still at 3s which I think is the main one people will action.

I did miss the snapshot timeout, I'm away from the computer for the night - would you mind adjusting this prior to the 9.0.2 release?

Supereg (approved):
As discussed I prepared the final state of the upcoming 0.9.2 release.

As discussed in private, HomeKit enforces a 10 second timeout on `/characteristics` requests anyways. Thus the our timeout values got reduced to fit into that time frame. Otherwise HomeKit will report all of our services as non responsive.

The Image snapshot handler was increased to have a ~ 22s timeout, as even HomeKit seems to be a bit more lenient with that.

I will merge this PR into beta for now and **draft** some release notes.
I await your approval, and not yet publish it as 0.9.2. Feel free to do that yourself if you are happy with those changes.

PR #890, opened 2021-03-08
Fix Validate Client Supplied Values + float min/max

oznu:
* Fix the `validateClientSuppliedValue` (renamed from `validClientSuppliedValue`)  method so it now transforms values to their corresponding types as expected (eg. transforms `false` to `0` for a UINT8 data type).
* Floats now have an upper and lower value defined so the validation methods used on INTs can largely be reused here.
* Check for non-finite numbers such as `Infinity` and `NaN` in client and user validation methods.
* When a `number` type is thrown or returned as a callback error, HAP will now verify it's a valid hap status number, or transform it into `HAPStatus.SERVICE_COMMUNICATION_FAILURE`.
* Added the `DEBUG_MESSAGE` level to `CharacteristicWarningType`
* Warnings about invalid write responses from a `SET` handler are now debug level characteristic warnings.

### DIFF
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -1963,7 +1963,7 @@ export class Characteristic extends EventEmitter {
           } // for stepValue < 1 rounding is done only when formatting the response. We can't store the "perfect" .step anyways
         }
 
-        break;
+        return value;
       }
       case Formats.STRING: {
         if (typeof value === "number") {
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -1579,7 +1579,7 @@ export class Characteristic extends EventEmitter {
           return this.value!;
         } else {
           if (writeResponse != null) {
-            this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`);
+            this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`, CharacteristicWarningType.DEBUG_MESSAGE);
           }
           this.value = value;
 
@@ -1642,7 +1642,7 @@ export class Characteristic extends EventEmitter {
               resolve(this.value!);
             } else {
               if (writeResponse != null) {
-                this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`);
+                this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`, CharacteristicWarningType.DEBUG_MESSAGE);
               }
               this.value = value;
               resolve();
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -1550,19 +1550,15 @@ export class Characteristic extends EventEmitter {
     // noinspection JSDeprecatedSymbols
     this.status = null;
 
-    if (connection !== undefined && !this.validClientSuppliedValue(value)) {
+    if (connection !== undefined) {
       // if connection is undefined, the set "request" comes from the setValue method.
       // for setValue a value of "null" is allowed and checked via validateUserInput.
-      return Promise.reject(HAPStatus.INVALID_VALUE_IN_REQUEST);
-    }
-
-    if (this.props.format === Formats.BOOL && typeof value === "number") {
-      // we already validate in validClientSuppliedValue that a number value can only be 0 or 1.
-      value = value === 1;
-    }
-
-    if (isNumericFormat(this.props.format) && typeof value === "boolean") {
-      value = value ? 1 : 0;
+      try {
+         value = this.validClientSuppliedValue(value);
+      } catch (e) {
+        debug(`[${this.displayName}]`, e.message);
+        return Promise.reject(HAPStatus.INVALID_VALUE_IN_REQUEST);
+      }
     }
 
     const oldValue = this.value;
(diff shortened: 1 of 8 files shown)


src/lib/util/request-util.ts
```diff
@@ -73,6 +73,8 @@ export function numericLowerBound(format: Formats | string): number {
   switch (format) {
     case Formats.INT:
       return -2147483648;
+    case Formats.FLOAT:
+      return Number.NEGATIVE_INFINITY;
```
Supereg, line 77: `numericLowerBound` and `numericUpperBound` return the "last" **valid** value for a given format right?
At least that's the case for all the other formats. So float is a bit off, and not really returning what you would expect from the function.
Returning +-infinity would mean we allow for float to be +- infinity?

src/lib/Characteristic.ts
```diff
@@ -1642,7 +1658,7 @@ export class Characteristic extends EventEmitter {
               resolve(this.value!);
             } else {
               if (writeResponse != null) {
-                this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`);
+                this.characteristicWarning(`SET handler returned write response value, though the characteristic doesn't support write response`, CharacteristicWarningType.DEBUG_MESSAGE);
```
Supereg, line 1661: Any particular reason to move this to debug level?
We don't want this to be user facing?

src/lib/Characteristic.ts
```diff
@@ -1742,135 +1758,100 @@ export class Characteristic extends EventEmitter {
...
+          } 
+          if (value > this.props.validValueRanges[1]) {
+            throw new Error(`Client supplied value of ${value} is greater than the maximum allowed value of ${this.props.validValueRanges[1]}`);
+          }
         }
 
-        numericMin = maxWithUndefined(this.props.minValue, numericLowerBound(Formats.UINT64));
-        numericMax = minWithUndefined(this.props.maxValue, numericUpperBound(Formats.UINT64));
-        break;
+        // should we enforce min step from the client as well?
```
Supereg, line 1816: Comment can be removed.
MinStep is only enforced when sending out the value, or rather when we format the value as a string.
We can't really store the rounded number precisely. And for the sake of preserving the "exact" number, rounding is only done when encoding values. 

Similar to the validateUserInput, we could however enforce that we don't have any floating point numbers for e.g. integer values.
So Enforce basically enforce the step value of `1` for integer values (basically enforcing that it is a integer type)

src/lib/Characteristic.ts
```diff
@@ -1915,158 +1896,82 @@ export class Characteristic extends EventEmitter {
...
-          this.characteristicWarning("characteristic value expected number and received " + typeof value);
-          value = typeof this.value === 'number' ? this.value : this.props.minValue || 0;
+
+        if (stepValue != undefined) {
+          if (stepValue === 1) {
+            value = Math.round(value);
+          } else if (stepValue > 1) {
+            value = Math.round(value);
+            value = value - (value % stepValue);
+          } // for stepValue < 1 rounding is done only when formatting the response. We can't store the "perfect" .step anyways
```
Supereg, line 1971: Just thinking about it maybe its worth removing the rounding here and to always store the exact value, only rounding the value inserted into the response.
Some plugins might not expect the value to be rounded? 

**except** for integer values we should enforce that they are indeed not a floating point value, meaning enforce a stepValue of at least `1`?

Would that make sense?

src/lib/Characteristic.ts
```diff
@@ -996,21 +996,27 @@ export class Characteristic extends EventEmitter {
...
         props.minValue = undefined;
-      } else if (isUnsignedNumericFormat(this.props.format)) {
+      } else if (typeof props.minValue !== 'number' || !Number.isFinite(props.minValue)) {
+        this.characteristicWarning(
+          `Characteristic Property 'minValue' must be a finite number, received "${props.minValue}" (${typeof props.minValue})`,
+          CharacteristicWarningType.ERROR_MESSAGE
+        );
+        props.minValue = undefined;
+      } else {
         if (props.minValue < numericLowerBound(this.props.format)) {
```
Supereg, line 1010: In conduction with the comment on numericLowerBound:
changing the code so it returns `-MAX_VALUE` instead of negative infinity, would ensure that this check also catches the case of `props.minValue=-infinity` (so we only need to check for NaN above).

Supereg (commented):
Thanks. Great work 👏

Had some smaller comments & questions here and there I feel like should be addressed before release. Other than that, looks great to me 👍

src/lib/Characteristic.ts
Shaquu: I don't feel like we need that to be user facing :)

Shaquu (dismissed):

src/lib/Characteristic.ts
oznu: Changing how the rounding works would be a change plugins may not be expecting in a patch release of Homebridge so I'd rather not change this logic just yet.

src/lib/util/request-util.ts
oznu: Changes to `-Number.MAX_VALUE` and `Number.MAX_VALUE`

src/lib/Characteristic.ts
oznu: Removed.

src/lib/Characteristic.ts
oznu: As per above changed to `Number.MAX_VALUE` etc. Keeping the `isFinite` number check in place is fine - as it's also catching NaN. If they try and set to `Infinity` or `NaN` then reverting to an unset maximum is better than guessing I think (not sure  how HomeKit is going to handle getting the number as `1.7976931348623157e+308`).

Supereg: You are right 👍 . Haven't thought about that.

Supereg (approved):

PR #902, opened 2021-08-11
Update Characteristic value when setProps called

Shaquu:
# Update Characteristic value when setProps called

This PR is mainly about number values.
Currently, Characteristic default value is based upon minValue.
Once setProps is called and minValue is changed, then also value should be changed accordingly.

## :bulb: Proposed solution
```typescript
public setProps(props: Partial<CharacteristicProps>): Characteristic {
  //...

  if (this.props.format) {
    switch (this.props.format) {
      case Formats.BOOL:
      case Formats.STRING:
      case Formats.DATA:
      case Formats.TLV8:
      case Formats.DICTIONARY:
      case Formats.ARRAY:
        break;
      default:
        if (this.value) {
          const minValue = this.props.minValue ?? 0
          const maxValue = this.props.maxValue ?? minValue
  
          if (this.value < minValue) {
            this.updateValue(minValue)
          } else if (this.value > maxValue) {
            this.updateValue(maxValue)
          }
        }
    }
  }	 

  return this;
}
```

### Problem that is solved
This PR will clamp Characteristic value, so there is no situation when allowed value range is 10-20 and value is 0.

### Implications
updateValue is called, which means some actions might be triggered

### Testing
Tested locally before and after with success.

### Reviewer Nudging
Right at the top :)


src/lib/Characteristic.ts
```diff
@@ -1143,7 +1143,30 @@ export class Characteristic extends EventEmitter {
...
+    if (this.props.format) {
+      switch (this.props.format) {
+        case Formats.BOOL:
+        case Formats.STRING:
+        case Formats.DATA:
+        case Formats.TLV8:
+        case Formats.DICTIONARY:
+        case Formats.ARRAY:
+          break;
+        default:
```
Supereg, line 1156: You might want to use `isNumericFormat` function defined in `request-util.ts`. This would reduce the required boilerplate a bit.

src/lib/Characteristic.ts
```diff
@@ -1143,7 +1143,30 @@ export class Characteristic extends EventEmitter {
...
+          if (this.value) {
+            const minValue = this.props.minValue ?? 0
+            const maxValue = this.props.maxValue ?? minValue
+
+            if (this.value < minValue) {
+              this.updateValue(minValue)
+            } else if (this.value > maxValue) {
+              this.updateValue(maxValue)
+            }
```
Supereg, line 1165: Thinking about it, checking for `minValue` and `maxValue` might not be enough. One can also specify numeric bounds using `validValues` and `validValueRanges` property.
Instead of reimplementing the whole logic, maybe you could reuse `validateUserInput`. This method is used e.g. in `setValue` to validate and sanitize any value coming from the user. Though one would probably need to silence any potential characteristic warnings when calling it from setProps.

Would be great if you could look into that, trying to reuse existing code as much as possible.

Supereg (commented):
Thanks for the PR. Great you caught that one 🚀 

My comments are mainly to direct you to reuse some of the already existent logic covering cases like that. Additionally, solely checking for the `minValue` and `maxValue` properties might not be enough to fully address the issue.

Shaquu:
@Supereg I will have a look at this soon, thanks for comments!

Shaquu:
Can you have a look @Supereg?
Also looks like there are some issues with tests on beta branch.

src/lib/Characteristic.ts
```diff
@@ -1180,6 +1180,9 @@ export class Characteristic extends EventEmitter {
       }
     }
 
+    // validateUserInput when called from setProps is intended to clamp value withing allowed range. It is why warnings should not be displayed.
+    this.updateValue(this.validateUserInput(this.value, CharacteristicWarningType.DEBUG_MESSAGE))
```
Supereg, line 1184: Way cleaner, thanks for addressing my comment 🚀

One thing I'm not sure how to handle: while `validateUserInput` typically can gracefully migrate the value to the updated properties, the method might still throw an error in fatal conditions. I'm not sure how to handle those situations. Especially in situations where we know the value won't be valid anymore (e.g. when changing the format). I think in those scenarios it might make sense to just init a new default value?

Furhter, setProps is called from the constructor, which makes the whole thing tricky on another level. At this point, the characteristic doesn't have a value yet (value is set to `null`). Also `updateValue` would be called even before the first value is assigned. I think, to avoid any complications, we should try to skip this line if we call `setProps` from the initializer. I think the most clean solution would be to define a private overload for `setProps` passing another parameter which controls execution of this check?

src/lib/Characteristic.ts
```diff
@@ -1896,9 +1899,10 @@ export class Characteristic extends EventEmitter {
    * in the future and throws an error which can't be converted to a valid value.
    *
    * @param value - The value received from the API call
+   * @param characteristicWarningType - Type of the warning message
    */
-  private validateUserInput(value?: Nullable<CharacteristicValue>): Nullable<CharacteristicValue> {
-    if (value === null) {
+  private validateUserInput(value?: Nullable<CharacteristicValue>, characteristicWarningType = CharacteristicWarningType.WARN_MESSAGE): Nullable<CharacteristicValue> {
+      if (value === null) {
```
Supereg, line 1905: nit pick, ident

src/lib/Characteristic.ts
```diff
@@ -1896,9 +1899,10 @@ export class Characteristic extends EventEmitter {
    * in the future and throws an error which can't be converted to a valid value.
    *
    * @param value - The value received from the API call
+   * @param characteristicWarningType - Type of the warning message
```
Supereg, line 1902: I would propose to be a bit more specific.

```suggestion
   * @param characteristicWarningType - Optionally defines the warning type to use when raising a {@link CharacteristicEventTypes.CHARACTERISTIC_WARNING}.
```

Supereg (commented):
Hey sorry for the delay, I wasn't sure if a follow-up review was already requested, and then kinda forgot about this 😕

Thanks for incorporating my feedback. I had some concerns on how `setProps` is used internally, as it is partly called in yet uninitialized state. Happy to year your points in that discussion :)

In regard to the tests, quickly scanned over it. Seems like, due to the added `validateUserInput` call (probably also because it is called through the constructor) the expected test results do not align anymore. Specifically the expectations on how many calls were made to `characteristicWarning` are now a bit off. Would be great if you could investigate those cases, and adjust the values if it makes sense.

src/lib/Characteristic.ts
Shaquu: Took some time... :)

I wonder if adding overload will not add another complexity.
Did you consider this topic?

Currently, if value is set to `null` then we would "fix it". Seems okay to me?

Shaquu:
@Supereg can I ask for hints or comments?

Shaquu:
@Supereg bump

Supereg (approved):
Looks good, thanks for the contribution 🚀

ebaauw:
Did you guys check this change for _Programmable Switch Event_ and other `eventOnly` characteristics, if there are any?  I fear this might cause a ghost button action.

I now get the “characteristic was supplied an illegal value” on restart, when having changed the supported button actions, and the last button action before the restart is no longer supported.  I thought about changing the characteristic value when changing the properties, but don’t know how to do that without generating a ghost button action.

Supereg:
My intention was to add some test cases in a follow up PR. Thanks for the input 👍 

> I now get the “characteristic was supplied an illegal value” on restart, when having changed the supported button actions, and the last button action before the restart is no longer supported.

You are justing calling `setProps` on a ProgrammabelSwitchEvent characteristic? Or am I misunderstanding something?

ebaauw:
Yes.  Based on the config.json settings, a button might support _Double Press_ and/or _Long Press_ in addition to _Single Press_.  E.g. the `doublePressTimeout` and `longPressTimeout` settings in Homebridge RPi, or the `hueDimmerRepeat` setting in Homebridge Hue.

On restart, Homebridge restores the _Stateless Programmable Switch_ from cache, including the last button action as the value of _Programmable Switch Event_.  My plugin sets the characteristic properties based on the changed config.json settings.  When the last button action from before the change is now no longer supported, Homebridge complains that the (restored) value is out of bound.

I think for `eventOnly` characteristics, there should be no check on `setProps()`, only on `setValue()`, `updateValue()`, and equivalents.  HomeKit won't ever read the characteristic, so it won't see the out of bound value anyways.

PR #912, opened 2021-11-24
Fix: Changed VOC[maxValue:5000], for support WirenBoard MSW sensors

Tehnologus:

* The system does not support a VOC value greater than 1000. We use WirenBoard MSW sensors, which shows VOC from 0 to 5000. Sensor information: https://wirenboard.com/wiki/WB-MSW_v.3_Modbus_Sensor/en

## :bulb: Proposed solution

* Change maxValue for VOC to 5000

### Testing

* Values in HomeKit on iPhone are shown correctly

### DIFF
--- a/src/lib/definitions/CharacteristicDefinitions.ts
+++ b/src/lib/definitions/CharacteristicDefinitions.ts
@@ -4846,7 +4846,7 @@ export class VOCDensity extends Characteristic {
       format: Formats.FLOAT,
       perms: [Perms.NOTIFY, Perms.PAIRED_READ],
       minValue: 0,
-      maxValue: 1000,
+      maxValue: 5000,
       minStep: 1,
     });
     this.value = this.getDefaultValue();


Shaquu:
I guess it is 1000 as it is specified like that in HAP Specification.
You can always change max value using props. No need to change source code.

Tehnologus:
> I guess it is 1000 as it is specified like that in HAP Specification. You can always change max value using props. No need to change source code.

"9.35 CurrentTemperature" is also  that must be greater than "zero" in HAP Specification.
But negative temperatures are shown normally

NorthernMan54:
> > I guess it is 1000 as it is specified like that in HAP Specification. You can always change max value using props. No need to change source code.
> 
> "9.35 CurrentTemperature" is also that must be greater than "zero" in HAP Specification. But negative temperatures are shown normally

And we change the minimum value using props in our code.

https://github.com/NorthernMan54/homebridge-tasmota/blob/87542fe850695317dfce2b7e550753c805d8dc55/src/tasmotaSensorService.ts#L50

Shaquu:
> > I guess it is 1000 as it is specified like that in HAP Specification. You can always change max value using props. No need to change source code.
> 
> 
> 
> "9.35 CurrentTemperature" is also  that must be greater than "zero" in HAP Specification.
> 
> But negative temperatures are shown normally
> 
> 

Do not feel discouraged.
Next time before you open PR is worth to ask around :) Maybe it can be solved without new release.

PR #918, opened 2021-12-30
New advertiser: Avahi/D-Bus API

adriancable:

HAP-NodeJS currently doesn't work on some set-ups ('No Response'!) due to some router bugs:

- Example: Arris G36 won't forward IPv4 multicast between wired and wireless interfaces. Bonjour-hap and Ciao advertisers won't advertise on IPv6. So you get no working advertisements if your HAP-NodeJS machine is connected via Ethernet instead of Wi-Fi.

HAP-NodeJS also currently doesn't work on other set-ups, for software-related reasons:

- Example: since on Linux Avahi is usually enabled by default, and HAP-NodeJS usually runs after Avahi, it cannot listen for unicast requests on 5353 even with SO_REUSEADDR. There are at least some routers which (even though they shouldn't) recreate multicast packets directed to a multicast group as unicast packets which are sent to connected devices joined to that group. Example: Arris RAC2V1A. So you get no working advertisements.

HAP-NodeJS a very long time ago used node-mdns as an advertiser which uses the Avahi Bonjour compatibility API, but:

- This required node-gyp to build as it used C/C++ modules, so it didn't work on some platforms
- The Bonjour compatibility API wasn't ever fully implemented, and throws up all sorts of warnings when you try to use it
- The Bonjour compatibility library was optional and some systems with otherwise complete Avahi don't have it

Also:

- Avahi has circa 10000X+ the user base of bonjour-hap/ciao and so probably has 10000X+ the hours of testing and therefore broader compatibility in general 

## :bulb: Proposed solution

Added a 3rd advertiser (in addition to, not replacing, bonjour-hap and ciao) which talks to Avahi over the D-Bus API. Because it's pure JavaScript (no node-gyp) it works on any system with Avahi/D-Bus. If there is no Avahi/D-Bus on the system just don't use it, and continue with ciao or bonjour-hap 'según el gusto'.

### Testing

Not done a lot. Seems to work on Linux! Tested IPv6 announcements, and they work.

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -1224,7 +1224,6 @@ export class Accessory extends EventEmitter {
         });
         break;
       case MDNSAdvertiser.AVAHI:
-        // noinspection JSDeprecatedSymbols
         this._advertiser = new AvahiAdvertiser(this._accessoryInfo);
         break;
       default:
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -29,7 +29,7 @@ import {
   VoidCallback,
   WithUUID,
 } from '../types';
-import { Advertiser, AdvertiserEvent, BonjourHAPAdvertiser, CiaoAdvertiser } from './Advertiser';
+import { Advertiser, AdvertiserEvent, BonjourHAPAdvertiser, CiaoAdvertiser, AvahiAdvertiser } from './Advertiser';
 // noinspection JSDeprecatedSymbols
 import { LegacyCameraSource, LegacyCameraSourceAdapter, StreamController } from './camera';
 import {
@@ -277,6 +277,10 @@ export const enum MDNSAdvertiser {
    * Use the `bonjour-hap` module as advertiser.
    */
   BONJOUR = "bonjour-hap",
+  /**
+   * Use Avahi/D-Bus as advertiser.
+   */
+  AVAHI = "avahi",
 }
 
 export type AccessoryCharacteristicChange = ServiceCharacteristicChange &  {
@@ -1218,6 +1222,10 @@ export class Accessory extends EventEmitter {
           restrictedAddresses: parsed.serviceRestrictedAddress,
           disabledIpv6: parsed.serviceDisableIpv6,
         });
+        break;
+      case MDNSAdvertiser.AVAHI:
+        // noinspection JSDeprecatedSymbols
+        this._advertiser = new AvahiAdvertiser(this._accessoryInfo);
         break;
       default:
         throw new Error("Unsupported advertiser setting: '" + info.advertiser + "'");
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -264,7 +264,7 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
 }
 
 /**
- * Advertiser based on the Avahi D-Bus library.
+ * Advertiser based on the Avahi D-Bus library. For (very crappy) docs on the interface, see the XML files at: https://github.com/lathiat/avahi/tree/master/avahi-daemon
  */
 export class AvahiAdvertiser extends EventEmitter implements Advertiser {
   private readonly accessoryInfo: AccessoryInfo;
@@ -274,7 +274,6 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
 
   private bus: any;
   private path: string = '';
-  private destroyed: boolean = false;
 
   constructor(accessoryInfo: AccessoryInfo) {
     super();
@@ -302,7 +301,6 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
   }
 
   public async startAdvertising(): Promise<void> {
-    assert(!this.destroyed, "Can't advertise on a destroyed Avahi instance!");
     if (this.port == undefined) {
       throw new Error("Tried starting Avahi advertisement without initializing port!");
     }
@@ -316,7 +314,6 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
 
   public async updateAdvertisement(silent?: boolean): Promise<void> {
     if (this.path) {
-      let txt = Object.entries(CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash)).map(el => Buffer.from(el[0] + '=' + el[1]));
       await this.avahiInvoke(this.path, 'EntryGroup', 'UpdateServiceTxt', { body: [-1, -1, 0, this.accessoryInfo.displayName, '_hap._tcp', '', this.createTxt()], signature: 'iiusssaay' });
     }
   }
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -12,6 +12,7 @@ import assert from "assert";
 import bonjour, { BonjourHAP, BonjourHAPService, MulticastOptions } from "bonjour-hap";
 import crypto from 'crypto';
 import createDebug from "debug";
+import dbus from 'dbus-native';
 import { EventEmitter } from "events";
 import { AccessoryInfo } from './model/AccessoryInfo';
 import { PromiseTimeout } from "./util/promise-utils";
@@ -262,3 +263,68 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
 
 }
 
+/**
+ * Advertiser based on the Avahi D-Bus library.
+ */
+export class AvahiAdvertiser extends EventEmitter implements Advertiser {
+  private readonly accessoryInfo: AccessoryInfo;
+  private readonly setupHash: string;
+
+  private port?: number;
+
+  private bus: any;
+  private path: string = '';
+  private destroyed: boolean = false;
+
+  constructor(accessoryInfo: AccessoryInfo) {
+    super();
+    this.accessoryInfo = accessoryInfo;
+    this.setupHash = CiaoAdvertiser.computeSetupHash(accessoryInfo);
+
+    this.bus = dbus.systemBus();
+
+    console.log(`Preparing Advertiser for '${this.accessoryInfo.displayName}' using Avahi backend!`);
+  }
+
+  private avahiInvoke(path: string, dbusInterface: string, member: string, others?: any): Promise<any> {
+    return new Promise((resolve, reject) => {
+      let command = { destination: 'org.freedesktop.Avahi', path, interface: 'org.freedesktop.Avahi.' + dbusInterface, member, ...(others || {}) };
+      this.bus.invoke(command, (err: any, result: any) => err ? reject(err) : resolve(result));
+    });
+  }
+
+  private createTxt(): Array<Buffer> {
+    return Object.entries(CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash)).map((el: Array<string>) => Buffer.from(el[0] + '=' + el[1]));
+  }
+
+  public initPort(port: number): void {
+    this.port = port;
+  }
+
+  public async startAdvertising(): Promise<void> {
+    assert(!this.destroyed, "Can't advertise on a destroyed Avahi instance!");
+    if (this.port == undefined) {
+      throw new Error("Tried starting Avahi advertisement without initializing port!");
+    }
+
+    console.log(`Starting to advertise '${this.accessoryInfo.displayName}' using Avahi backend!`);
+
+    this.path = await this.avahiInvoke('/', 'Server', 'EntryGroupNew') as string;
+    await this.avahiInvoke(this.path, 'EntryGroup', 'AddService', { body: [-1, -1, 0, this.accessoryInfo.displayName, '_hap._tcp', '', '', this.port, this.createTxt()], signature: 'iiussssqaay' });
+    await this.avahiInvoke(this.path, 'EntryGroup', 'Commit');
+  }
+
+  public async updateAdvertisement(silent?: boolean): Promise<void> {
+    if (this.path) {
+      let txt = Object.entries(CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash)).map(el => Buffer.from(el[0] + '=' + el[1]));
+      await this.avahiInvoke(this.path, 'EntryGroup', 'UpdateServiceTxt', { body: [-1, -1, 0, this.accessoryInfo.displayName, '_hap._tcp', '', this.createTxt()], signature: 'iiusssaay' });
+    }
+  }
+
+  public async destroy(): Promise<void> {
+    if (this.path) {
+      await this.avahiInvoke(this.path, 'EntryGroup', 'Free');
+      this.path = '';
+    }
+  }
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@homebridge/ciao": "~1.1.3",
         "bonjour-hap": "~3.6.3",
+        "dbus-native": "^0.4.0",
         "debug": "^4.3.2",
         "fast-srp-hap": "2.0.4",
         "futoin-hkdf": "~1.4.2",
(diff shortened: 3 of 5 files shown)


src/lib/Advertiser.ts
```diff
@@ -262,3 +263,68 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
...
+  private createTxt(): Array<Buffer> {
+    return Object.entries(CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash)).map((el: Array<string>) => Buffer.from(el[0] + '=' + el[1]));
+  }
+
+  public initPort(port: number): void {
+    this.port = port;
+  }
+
+  public async startAdvertising(): Promise<void> {
+    assert(!this.destroyed, "Can't advertise on a destroyed Avahi instance!");
```
Supereg, line 305: `destroyed` property is never set.

src/lib/Accessory.ts
```diff
@@ -1219,6 +1223,10 @@ export class Accessory extends EventEmitter {
           disabledIpv6: parsed.serviceDisableIpv6,
         });
         break;
+      case MDNSAdvertiser.AVAHI:
+        // noinspection JSDeprecatedSymbols
```
Supereg, line 1227: I think this command can be removed.

src/lib/Advertiser.ts
```diff
@@ -262,3 +263,68 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
...
+  public async updateAdvertisement(silent?: boolean): Promise<void> {
+    if (this.path) {
+      let txt = Object.entries(CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash)).map(el => Buffer.from(el[0] + '=' + el[1]));
+      await this.avahiInvoke(this.path, 'EntryGroup', 'UpdateServiceTxt', { body: [-1, -1, 0, this.accessoryInfo.displayName, '_hap._tcp', '', this.createTxt()], signature: 'iiusssaay' });
+    }
+  }
+
+  public async destroy(): Promise<void> {
+    if (this.path) {
+      await this.avahiInvoke(this.path, 'EntryGroup', 'Free');
```
Supereg, line 326: I'm not entirely familiar with the dbus interface. Will the advertisement be removed when the application exists non-gracefully, or will the advertisement turn into an orphan?

src/lib/Advertiser.ts
```diff
@@ -262,3 +263,68 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
 
 }
 
+/**
+ * Advertiser based on the Avahi D-Bus library.
```
Supereg, line 267: Would you mind linking to some resource where the D-Bus interface of avahi is documented?

Supereg (commented):
Thanks for you contribution. Especially for taking the time to illustrate the situation around mdns advertising with HAP-NodeJS. 🚀 

I very much like the idea of getting back to interface with "standard" mdns implementations. This was and is a reoccurring pain point of HAP-NodeJS. Having a way to interface with avahi in a non-complex way is definitely a huge win 🎉 

I know this is not part of the PR, but is there an easy way of checking if dbus and avahi is available on the currently running platform?

On a similar topic, I already though about interfacing with `mDNSResponder` on macOS. To my knowledge, the easiest approach would be to interface via the `dns-sd` CLI. Though, doing txt updates isn't really possible without completely readvertising, but that might be a sensible addition for the future as well.

src/lib/Advertiser.ts
adriancable: Oops! I'll remove this assert. Unlike with the other advertisers, it's fine to destroy and then startAdvertising again after if you wish with Avahi/D-Bus.

src/lib/Accessory.ts
adriancable: OK!

src/lib/Advertiser.ts
adriancable: Yes, that's right. Communication with D-Bus is over a fd which gets closed when the process exits for any reason. Avahi automatically removes the advertisement when the fd is closed.

src/lib/Advertiser.ts
adriancable: I'll add a link to what I used, which is the XML files in the Avahi repository. I couldn't find any decent docs for Avahi's D-Bus interface, so it was a bit of trial and error.

adriancable:
> I know this is not part of the PR, but is there an easy way of checking if dbus and avahi is available on the currently running platform?

Yes. If D-Bus is not available, then this line will throw an exception (tested on my Mac which *can* have D-Bus but does not have it installed):

`this.bus = dbus.systemBus();`

I can easily catch this, but not sure what the right way of dealing with it is. (in the sense of what would "fit in" best with the rest of HAP-NodeJS's operation)

If D-Bus is available but Avahi is not, then the avahiInvoke calls *should* throw an exception. (hard for me to test this)

src/lib/Advertiser.ts
Supereg: Ah perfect 👍 Good to hear that this is reliable.

src/lib/Advertiser.ts
Supereg: Thanks. 🎉

Supereg:
> > I know this is not part of the PR, but is there an easy way of checking if dbus and avahi is available on the currently running platform?
> 
> Yes. If D-Bus is not available, then this line will throw an exception (tested on my Mac which _can_ have D-Bus but does not have it installed):
> 
> `this.bus = dbus.systemBus();`
> 
> I can easily catch this, but not sure what the right way of dealing with it is. (in the sense of what would "fit in" best with the rest of HAP-NodeJS's operation)
> 
> If D-Bus is available but Avahi is not, then the avahiInvoke calls _should_ throw an exception. (hard for me to test this)

Okay good to know. I think it is fine if we keep the "auto configuration" separate from this PR for now.

> If D-Bus is available but Avahi is not, then the avahiInvoke calls _should_ throw an exception. (hard for me to test this)

I think the easiest way to test this, is to boot into a barebones docker container. Those typically ship without avahi.

adriancable:
> I think the easiest way to test this, is to boot into a barebones docker container. Those typically ship without avahi.

Even easier is probably just to `systemctl stop avahi-daemon`!

Supereg (approved):
🚀

Supereg:
The PR is live with 0.10.0-beta.4

ebaauw:
Thank you both!  I’ll be testing this on my Raspberry Pi installations.  Looking good so far (in Discovery app) on a test installation with five child bridges.

ebaauw:
Hm, getting a fatal error on Raspberry Pi Zero:
```
Dec 31 06:14:40] [31/12/2021, 06:14:40] [HB Supervisor] Restarting Homebridge...
[Dec 31 06:14:40] [31/12/2021, 06:14:40] [HB Supervisor] Starting Homebridge with extra flags: -T -D
[Dec 31 06:14:40] [31/12/2021, 06:14:40] [HB Supervisor] Starting Homebridge with custom env: {"NODE_OPTIONS":"--trace-warnings"}
[Dec 31 06:14:40] [31/12/2021, 06:14:40] [HB Supervisor] Started Homebridge v1.4.0-beta.1 with PID: 32243
[Dec 31 06:14:56] #
[Dec 31 06:14:56] # Fatal error in , line 0
[Dec 31 06:14:56] # Liftoff bailout should not happen. Cause: Armv6 not supported
[Dec 31 06:14:56] #
[Dec 31 06:14:56] #
[Dec 31 06:14:56] #
[Dec 31 06:14:56] #FailureMessage Object: 0xb49faf50
[Dec 31 06:14:57] [31/12/2021, 06:14:57] [HB Supervisor] Homebridge Process Ended. Code: null, Signal: SIGTRAP
```
I also though I saw `node-gyp` being invoked while installing the Homebridge beta?

donavanbecker:
> The PR is live with 0.10.0-beta.4
> 
> EDIT: live with homebridge v1.4.0-beta.0. The UI still needs to be updated to be able to configure the advertiser.

@oznu

donavanbecker:
getting the following on M1 Mac:

```
Error: connect ENOENT /var/run/dbus/system_bus_socket
    at PipeConnectWrap.afterConnect [as oncomplete] (node:net:1161:16)
```

ebaauw:
> I also though I saw `node-gyp` being invoked while installing the Homebridge beta?

Yes, looking at `htop`, while installing `es-abstract` after installing `hap-nodejs`.  I don’t think `npm` saved a log file, as the install (eventually) succeeds.  This is not the case for installing the `latest` version of Homebridge.

> Because it's pure JavaScript (no node-gyp) it works on any system with Avahi/D-Bus.

Apparently not?

ebaauw:
> getting the following on M1 Mac

That’s to be expected.  No dbus nor Avahi on Mac.  I hope only when setting `advertiser` to `avahi`?

donavanbecker:
> > getting the following on M1 Mac
> 
> That’s to be expected. No dbus nor Avahi on Mac. I hope only when setting `advertiser` to `avahi`?

yes.

adriancable:
> > Because it's pure JavaScript (no node-gyp) it works on any system with Avahi/D-Bus.
> 
> Apparently not?

`dbus-native` uses `abstract-socket` as an `optionalDependency`, which does use node-gyp. But as far as I can tell this is just to improve performance. If node-gyp isn't found or fails, everything still works. Alternatively, you can skip installing it using npm i --no-optional, node-gyp isn't called, and everything also still works.

oznu:
Looks good. I'm no where near a computer right not to actually test however.

Before this goes into a Homebridge releases I'd like to see auto-fallback to a valid advertiser if Avahi is not available on the system (assuming this is not already the case?), and the default to remain as Bonjour for existing installs for at least for the initial release, (I can make it so new installs have the advertiser set to Avahi if available).

adriancable:
> [Dec 31 06:14:56] # Liftoff bailout should not happen. Cause: Armv6 not supported

Are you using Node 17? ARMv6 is no longer fully supported. Please go back to Node 16 in this case.

ebaauw:
Node v16.13.1.  And `sudo npm -g i homebridge@beta --no-optional` still calls `node-gyp`.

adriancable:
> Node v16.13.1. And `sudo npm -g i homebridge@beta --no-optional` still calls `node-gyp`.

Try node --no-expose-wasm

Regarding calling node-gyp, which package is calling it? If it's `es-abstract` (you mentioned this earlier), that's a dependency of deep-equal which is a dependency of bonjour-hap, so not related to this PR.

ebaauw:
> Try node --no-expose-wasm

Running Homebridge or running `npm`?

> Regarding calling node-gyp, which package is calling it? If it's es-abstract (you mentioned this earlier), that's a dependency of deep-equal which is a dependency of bonjour-hap, so not related to this PR.

Not sure.  How can I force `npm` to keep a log on success.  It’s a dependency that’s introduced in the new beta, as `node-gyp` isn’t invoked when installing the `latest` version.

ebaauw:
On a positive note, the avahi/dbus advertiser seems to work fine on my Pi 4B servers, running 32-bit buster or bullseye.  I have the impression that Home on my iPad spends a little less time on updating the tiles than when using ciao.

Supereg:
> Looks good. I'm no where near a computer right not to actually test however.
> 
> Before this goes into a Homebridge releases I'd like to see auto-fallback to a valid advertiser if Avahi is not available on the system (assuming this is not already the case?), and the default to remain as Bonjour for existing installs for at least for the initial release, (I can make it so new installs have the advertiser set to Avahi if available).

@oznu as of now bonjour-hap is still the default and avahi has to be manually configured. I would give it a shot to implement the auto-default in hap-nodes next week. Not sure if this needs modifications then in the UI (apart from adding a new case) to properly display the default in the dropdown.

adriancable:
> > Try node --no-expose-wasm
> 
> Running Homebridge or running `npm`?

Running whatever it was that gave the ARMv6 error.

adriancable:
> **Edit** found the log, indeed it’s `abstract-socket`:

From some Googling, it appears that `npm --no-optional` is ignored if the package-lock.json was made without `--no-optional`. So this is probably why you're still seeing node-gyp running. Can you try deleting package-lock.json and then running `npm i --no-optional` again?

Also `npm i` will still succeed even if an optionalDependency can't install because of lack of node-gyp or similar. So I still think the way things are right now is fine ... `npm i` will still work to install HAP-NodeJS even if abstract-socket isn't supported on the platform. And, if it's a platform where abstract-socket isn't supported, probably Avahi/D-Bus isn't supported either, so in practice no issue.

ebaauw:
The issue is that the Homebridge beta crashes on a Pi Zero, before it’s even read config.json to find out which advertiser to use.  I only use the Pi Zero for testing, so I couldn’t care less, but there are people out there actually running Homebridge on an arm6.  This needs to be fixed before publishing a released version with the Avahi/D-Bus advertiser, or support for the arm6 needs to be dropped officially - again, I don’t care which; I’m just reporting what I encounter.  This is a standard installation as per the Wiki, and and I installed the beta through the UI.  I’m willing and able to try stuff on the command line, in order to analyse this issue, but this _is_ a real issue, in practice.

adriancable:
> The issue is that the Homebridge beta crashes on a Pi Zero, before it’s even read config.json to find out which advertiser to use. I only use the Pi Zero for testing, so I couldn’t care less, but there are people out there actually running Homebridge on an arm6. This needs to be fixed before publishing a released version with the Avahi/D-Bus advertiser, or support for the arm6 needs to be dropped officially - again, I don’t care which; I’m just reporting what I encounter. This is a standard installation as per the Wiki, and and I installed the beta through the UI. I’m willing and able to try stuff on the command line, in order to analyse this issue, but this _is_ a real issue, in practice.

I understand, but it's not clear how you came to the conclusion that the ARMv6 crash has something to do with the addition of the Avahi/D-Bus advertiser, and not some other dependency that's been updated in the Homebridge beta. In fact since your issue seems to come from using Homebridge, it isn't even clear that it's related to HAP-NodeJS at all. Can you share what info you have on this? Thanks!

bwp91:
> support for the arm6 needs to be dropped officially

i don’t think this could happen until at least april 2023 when homebridge *officially* drops support for node 14. unless i am getting my versions of node that support arm6 mixed up!

Supereg:
@ebaauw @adriancable Would be great if you could track it in a new, separate issue (tagged with `beta`).

ebaauw:
> but it's not clear how you came to the conclusion that the ARMv6 crash has something to do with the addition of the Avahi/D-Bus advertiser

I haven't.  I'm just reacting here as Andi mentioned the Homebridge beta here.

> Can you share what info you have on this?

I'm afaid, I already have.  Frankly, I have no idea who issues the messages, not who causes the SIGTRAP.  I'm not even sure if it's related to the invocation of `node-gyp`.  This is just another discrepancy I noticed.  Attached the `npm` log when installing the beta.
[2021-12-31T07_15_18_070Z-debug-0.log](https://github.com/homebridge/HAP-NodeJS/files/7796822/2021-12-31T07_15_18_070Z-debug-0.log)

ebaauw:
> Would be great if you could track it in a new, separate issue (tagged with beta).

In HAP-NodeJS or in Homebridge?

Supereg:
> > Would be great if you could track it in a new, separate issue (tagged with beta).
> 
> In HAP-NodeJS or in Homebridge?

HAP-NodeJS. Thanks!

ebaauw:
> Would be great if you could track it in a new, separate issue (tagged with beta).

@Supereg : https://github.com/homebridge/HAP-NodeJS/issues/919

> Try node --no-expose-wasm

@adriancable : That does the trick, see the issue. Does that mean the crash is related to `dbus-native`?

Again, on a positive note: I'm very happy with the PR, and it seems to run perfectly well on the Pi 4B.  As I mentioned above, I have a feeling Home is spending less time _Updating_ the tiles.

adriancable:
> > Try node --no-expose-wasm
> 
> @adriancable : That does the trick, see the issue. Does that mean the crash is related to `dbus-native`?

No, unfortunately, I don't have any ARMv6 platform to test. But Googling around, I read that Node is deprecating ARMv6 from 16.1 onwards, and at least for now this command line switch seems to disable the bits which break.

adriancable:
> Again, on a positive note: I'm very happy with the PR, and it seems to run perfectly well on the Pi 4B. As I mentioned above, I have a feeling Home is spending less time _Updating_ the tiles.

I also (independently) noticed the exact same thing. I am fairly sure the effect is real and not placebo. I wonder why this would be the case?

PR #904, opened 2021-08-28
HomeKit Secure Video

koush:
HomeKit Secure Video implementation.

Spec:
https://github.com/Supereg/secure-video-specification

## Demo/Testing

It uses Scrypted instead of Homebridge:

https://github.com/koush/scrypted

Then install one of the Scrypted Unifi, Amcrest, Hikvision, or Onvif plugins via the web console. You may also want to install the Rebroadcast plugin, it retains a video loop for HSKV that can be quickly served up on motion. It's in the spec, but seems to be a soft requirement. Not sending the prebuffer results in a poor experience as you may miss what triggered the motion.

## Explanation

I won't get into the spec, you can read that. Basically HAP-NodeJS was modified to request the recording options available for a device and create the appropriate RecordingManagement that communicates with HomeKit. When motion is triggered, HomeKit dials in to open a data connection with HAP-NodeJS. HAP-NodeJS will invoke a callback for the plugin to generate mp4 fragments. Creating those fragments properly is up to the plugin. HAP-NodeJS will then properly chunk and stream the fragments to the Home hub.

## Debugging HAP-NodeJS:

```sh
# assuming server is already running
cd scrypted/plugins/homekit
npm install
# launch vs code
code .
# Press the play button to install the plugin and launch the debugger
```

## My Setup

I've got 9 cameras running this, works great. HomePods is more reliable than Apple TV which for whatever reason stops responding to motion events and triggering recordings until restarted. Unclear why, but I noticed this only happens with Apple TV.

Screen Shot 2021-08-27 at 8 56 39 PM

PR #920, opened 2022-01-03
HomeKit Secure Video

Supereg:

PR #904 originally proposed to implement HomeKit Secure Video support in HAP-NodeJS. The PR of the original author is considered stale since a few weeks now.

## :bulb: Proposed solution

This PR replaces #904 and realizes the requested changes, building upon the existing PR and moving it from a rather experimental state to a state where the API can be considered stable. The implementation itself was extended to accommodate for much more extensive error and edge case handling.

To highlight some advancements to the previous PR (from a API interface perspective):
* The `CameraControllerOptions` were refactored:
  * Added detailed documentation to all fields
  * `EventTriggerOption`s are now derived automatically based on the `CameraController` or `DoorbellController` configuration
  * Added proper typing to all fields
  * A `MotionSensor` (and a `OccupancySensor`) can be either be enabled by flag or passed as an instance. The second option can be used to support people in migrating from preexisting MotionSensor instances to managing them with the CameraController.
* Rethought the `CameraRecordingDelegate` interface:
  * Added documentation.
  * Introduces calls to be notified of changes to the `SelectedRecordingConfiguration` and the `Active` state. Second one is used to enable or disable the prebuffer.
  * Added dedicated calls to close the streaming request. Previously this was done through `AsyncGenerator.throw` which wasn't ideal as (a) one couldn't transport additional context or information, (b) it is a feature which isn't really intuitive nor documented well, (c) it is slow, as the generator is only notified on the next `yield` and (d) it is easy to make mistakes, resulting in a potential application crash if done wrong. The new API calls are more intuitive and allow to instantly abort the generator.
  * Properly enforce a single recording stream at a time.
  * Added support to mark a recording stream as `endOfStream`.
* SnapshotRequest are now checked if they are actually allowed in the current state (Previously this was expected to be done by every individual developer).
* Implemented state handling. The HSV camera will save and restore its configuration state across reboots.
* Fixed some bugs in service and state handling.

## :gear: Release Notes

*This section is addressed once merged, such that we can link to docs pages.*

* Short spec writeup: https://github.com/Supereg/secure-video-specification. Though inline docs can be considered more detailed and contain a bit newer information from my latest discoveries and tests.

### Testing

The PR adds some minimal unit tests for the `CameraController`.
Further, it implements a minimal `CameraRecordingDelegate` in the `IPCamera_accessory.ts` example, providing a rough overview on how to use the API. But most importantly this serves as a test vector to test against Apples HomeKit Controller implementation.

### Reviewer Nudging

Also applying to users interesting in implementing HomeKit Secure Video.
In any case I would propose that you carefully read through the inline documentation provided with user facing interfaces. Additionally, have a look at the example HAP-NodeJS accessory implementation, giving a minimal but working recording delegate.


Casuallynoted (dismissed):

seydx:
Works great. Thx @Supereg 

**Possibly worth mentioning:**

- If already existing cameras in Apple Home now become HSV capable, it may happen that they do not record immediately. Turning "**Stream & Recording**" within Apple Home off and on several times seems to solve the problem. If that doesn't help, restart the hub.

- If a movement is triggered "**manually**" (e.g. via a switch), don't be surprised that you don't get a notification. HSV seems to analyze the recording and only if there really is motion, you get a notification or you can see the recording.

- If **vcodec = copy** is used via FFmpeg, you should make sure that the camera settings are suitable for HSV.  FPS should be > 20 and Key Frame Interval = FPS x 4 (e.g. 20 x 4 = 80). 

### My camera settings:
**FPS:** 25
**Key Frame Interval:** 100
**Resolution:** 1920x1080
**Bitrate:** 2048 kbps (CBR)

My implementation:

https://github.com/seydx/homebridge-camera-ui/blob/hsv/src/accessories/camera.js
https://github.com/seydx/homebridge-camera-ui/blob/hsv/src/services/recording.service.js

Supereg:
Maybe to give some context to these points:

>  If already existing cameras in Apple Home now become HSV capable, it may happen that they do not record immediately.

Recording of newly added HSV camera is always turned **off by default** (same behavior as for certified cameras). So the step of manually enabling recording is expected and required.

> HSV seems to analyze the recording (?) and only if there really is motion, you get a notification or you can see the recording.

Every HSV camera has two major options controlling these behaviors. The `recording options` (which only appear if recording is enabled) which control when a recording is **saved** (e.g. checking for people, animals etc; also consider that activity zones might influence this behavior). And secondly you have the `Activity Notifications` setting in your `Notifications` settings which is **off by default**. With this toggle you can selected to be notified when `A Clip is Recorded`, `Any Motion is Detected` or `Specific Motion is Detected` (last options ties back to the recording options).
I would assume your findings are probably related to those options?

> If vcodec = copy is used via FFmpeg, you should make sure that the camera settings are suitable for HSV. FPS should be > 20 and Key Frame Interval = FPS x 4 (e.g. 20 x 4 = 80).

This is really dependent on the camera. In general, it is noted advised to use `vcodec=copy` as your cameras video configuration is most likely different to the recording options HomeKit selects and **expects**. What you are seeing here is the enforcement of "every fragment starts with a key frame" (assuming you have a fragment size of 4000ms). Therefore you only reach this requirement (considering your fixed key fragment length) with your fixed camera with a fps count of 20. This obviously doesn't hold for everyone.

PR #922, opened 2022-01-06
Automatically use Avahi/DBus advertiser if available on platform

Supereg:

PR #918 introduced a new advertiser using native avahi over its dbus interface. The advertiser required to be manually enabled.

## :bulb: Proposed solution

This PR adds additional heuristic to check if (a) dbus is installed and (b) avahi is installed. If both is the case hap-nodejs will use the `AVAHI` advertiser if nothing else is specified.

## :gear: Release Notes

* The new `avahi` based advertiser will be automatically used on supported platforms.

Note, due to the fact that the new checks are performed asynchronously, the `Accessory.publish` signature was changed to return a `Promise<void>` instead of `void`.

### Testing

Added an additional test case, ensuring that the availability check succeeds in CI environment.

### Reviewer Nudging

--

### DIFF
--- a/src/lib/Accessory.spec.ts
+++ b/src/lib/Accessory.spec.ts
@@ -131,7 +131,7 @@ describe('Accessory', () => {
         advertiser: advertiser
       };
 
-      accessory.publish(publishInfo);
+      await accessory.publish(publishInfo);
 
       expect(accessory.displayName.startsWith(DEFAULT_DISPLAY_NAME));
       expect(accessory.displayName.length).toEqual(DEFAULT_DISPLAY_NAME.length + 1 + 4); // added hash!
@@ -149,7 +149,7 @@ describe('Accessory', () => {
 
       await PromiseTimeout(200);
 
-      accessory.publish(publishInfo)
+      await accessory.publish(publishInfo)
 
       // ensure unification isn't done twice!
       expect(accessory.displayName).toEqual(displayNameWithIdentifyingMaterial);
@@ -157,7 +157,30 @@ describe('Accessory', () => {
       await awaitEvent(accessory, AccessoryEventTypes.ADVERTISED);
 
       await accessory.unpublish();
-    })
+    });
+
+    test("Clean Accessory publish and unpublish with default advertiser selection", async () => {
+      const accessory = new Accessory(DEFAULT_DISPLAY_NAME, uuid.generate("foo"));
+
+      const switchService = new Service.Switch("My Example Switch");
+      accessory.addService(switchService);
+
+      let publishInfo: PublishInfo = {
+        username: TEST_USERNAME,
+        pincode: "000-00-000",
+        category: Categories.SWITCH,
+        advertiser: undefined,
+      };
+
+      await accessory.publish(publishInfo);
+
+      expect(accessory.displayName.startsWith(DEFAULT_DISPLAY_NAME));
+      expect(accessory.displayName.length).toEqual(DEFAULT_DISPLAY_NAME.length + 1 + 4); // added hash!
+
+      await awaitEvent(accessory, AccessoryEventTypes.ADVERTISED);
+
+      await accessory.unpublish();
+    });
   });
 
   describe("characteristicWarning", () => {
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -1109,7 +1109,7 @@ export class Accessory extends EventEmitter {
    *                                that for instance an appropriate icon can be drawn for the user while adding a
    *                                new Accessory.
    */
-  public publish(info: PublishInfo, allowInsecureRequest?: boolean): void {
+  public async publish(info: PublishInfo, allowInsecureRequest?: boolean): Promise<void> {
     // noinspection JSDeprecatedSymbols
     if (!info.advertiser && info.useLegacyAdvertiser != null) {
       // noinspection JSDeprecatedSymbols
@@ -1206,7 +1206,10 @@ export class Accessory extends EventEmitter {
     // create our Advertiser which broadcasts our presence over mdns
     const parsed = Accessory.parseBindOption(info);
 
-    switch (info.advertiser ?? MDNSAdvertiser.BONJOUR) {
+    const defaultAdvertiser = info.advertiser
+      ?? (await AvahiAdvertiser.isAvailable() ? MDNSAdvertiser.AVAHI : MDNSAdvertiser.BONJOUR);
+
+    switch (defaultAdvertiser) {
       case MDNSAdvertiser.CIAO:
         this._advertiser = new CiaoAdvertiser(this._accessoryInfo, {
           interface: parsed.advertiserAddress
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -1325,7 +1325,8 @@ export class Accessory extends EventEmitter {
     this._advertiser!.initPort(port);
 
     this._advertiser!.startAdvertising()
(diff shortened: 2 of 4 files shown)


Supereg:
@adriancable I noticed the following shortcomings with the current avahi advertiser implementation (including what's done in this PR).

* The `isAvailable` check will only check if avahi is installed. Meaning it will also succeed if avahi is stopped. And actually it will lead to avahi getting started. See:
```
Jan 06 18:50:57 raspberrypi3 dbus-daemon[349]: [system] Activating via systemd: service name='org.freedesktop.Avahi' unit='dbus-org.freedesktop.Avahi.service' requested by ':1.195' (uid=1000 pid=10177 comm="node /usr/bin/ts-node --files src/BridgedCore.ts ")
Jan 06 18:50:57 raspberrypi3 systemd[1]: Listening on Avahi mDNS/DNS-SD Stack Activation Socket.
Jan 06 18:50:57 raspberrypi3 systemd[1]: Starting Avahi mDNS/DNS-SD Stack...
```
I have tried to use difference methods; e.g. the [`GetState`](https://github.com/lathiat/avahi/blob/fd482a74625b8db8547b8cfca3ee3d3c6c721423/avahi-daemon/org.freedesktop.Avahi.Server.xml#L59) (see [here](https://github.com/lathiat/avahi/blob/fd482a74625b8db8547b8cfca3ee3d3c6c721423/avahi-common/defs.h#L221-L227) for values). Sounded promising at the beginning, but this will also lead to avahi getting started. 
* The current advertiser does not detect if avahi is stopped or rebooting. This has two effects (a) our service record won't be advertised anymore until we restart hap-nodejs and (b) somewhere this causes a `UnhandledPromiseRejectionWarning` on shutdown. This issue is tracked in #923.

Supereg:
apart from those issues, something seems to be failing in the CI environment such that this thing runs forever 🙃

adriancable:
> * The `isAvailable` check will only check if avahi is installed. Meaning it will also succeed if avahi is stopped. And actually it will lead to avahi getting started. See:
> 
> * The current advertiser does not detect if avahi is stopped or rebooting. This has two effects (a) our service record won't be advertised anymore until we restart hap-nodejs and (b) somewhere this causes a `UnhandledPromiseRejectionWarning` on shutdown. This issue is tracked in [Avahi/DBus: Shutdown of avahi stays undetected and leads to UnhandledPromiseRejection on shutdown. #923](https://github.com/homebridge/HAP-NodeJS/issues/923).

Hi @Supereg - I have been thinking about this and it isn't clear to me that either of these things are wrong.

To your 1st point, I think if Avahi is stopped and the user then requests Avahi as an advertiser in Homebridge, it's the right thing for it to get started. So I think the behaviour is correct. Why do you need to distinguish the case 'a priori' whether Avahi is running or stopped? (If you do want that, you can just do `pidof avahi-daemon` but I'm not sure why that would be needed at all.)

To your 2nd point, Avahi is always installed as a service which means by default it's running. So Avahi will be only stopped if the user manually stops it. In this case the user must expect that all Avahi services would stop being advertised! So I think what happens now is correct.

Supereg:
> To your 1st point, I think if Avahi is stopped and the user then requests Avahi as an advertiser in Homebridge, it's the right thing for it to get started. So I think the behaviour is correct. Why do you need to distinguish the case 'a priori' whether Avahi is running or stopped? (If you do want that, you can just do pidof avahi-daemon but I'm not sure why that would be needed at all.)

I completely agree. Just found it to be bit weird that this just automatically starts avahi. Not sure what e.g. a users decision might be to have avahi stopped. Just felt like something you won't expect.
But in any case. Less handling for us I guess.

> To your 2nd point, Avahi is always installed as a service which means by default it's running. So Avahi would only be stopped if the user manually stops it. In this case the user must expect that all Avahi services would stop being advertised! So I think what happens now is correct.

My main point here really is, that one would expect that the service re-registers itself automatically. Ideally we could handle that (but seemingly we can't. Didn't see any signals coming in). Otherwise I would definitely document this.

Supereg:
PR ready for review.

* Fixed to CI hang, as jest wouldn't exist due to the still open message bus socket.
* Added some error handling to handle the above edge cases where e.g. cleanup might fail when avahi responser was stopped.

NorthernMan54 (approved):

NorthernMan54:
@Supereg A couple of comments from my testing

1 - On OSX - I couldn't get HAP-NodeJS to compile under typescript, ran into what looked like this - https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44946, so no test results.  Spent a bit of time trying things out, but got stuck.

2 - On my main RPI, complied okay, which was odd as it is using v14.18.2 of node and my OSX was using v16.13.1.  Then delved into why when running under Homebridge it kept defaulting to bonjour.  Ended up hacking out the current homebridge behaviour where if an advertiser is not declared it defaults to bonjour.

After hacking homebridge, it worked correctly and it selected Avahi

```
  HAP-NodeJS:Advertiser Detected Avahi over DBus interface running version 'avahi 0.7'. +0ms
  HAP-NodeJS:Advertiser Preparing Advertiser for 'Leonard DD97' using Avahi backend! +26ms
  HAP-NodeJS:Advertiser Starting to advertise 'Leonard DD97' using Avahi backend! +31ms
```

This particular RPI is a bit back level on raspbian and currently has 

```
cat /etc/os-release 
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
```

After looking at how the Homebridge default advertiser logic works, this may need some further thought as how we need or want to behave.

Supereg:
> 1 - On OSX - I couldn't get HAP-NodeJS to compile under typescript, ran into what looked like this - DefinitelyTyped/DefinitelyTyped#44946, so no test results. Spent a bit of time trying things out, but got stuck.

Could you supply the exact error message? Does this only happen on this feature branch or also on the main branch? In any case, would be great if you could open an issue with the exact error message when trying to compile.

> Ended up hacking out the current homebridge behaviour where if an advertiser is not declared it defaults to bonjour.

Yeah, sorry for not clarifying this. Homebridge wasn't updated for this feature yet. This PR is the first step.

> After looking at how the Homebridge default advertiser logic works, this may need some further thought as to how we need or want it behave.

Are you referring to the current **homebridge** default advertiser selection? This always defaults to the `bonjour` package and isn't updated to allow `hap-nodejs` to run its own default advertiser selection.

NorthernMan54:
> > 1 - On OSX - I couldn't get HAP-NodeJS to compile under typescript, ran into what looked like this - [DefinitelyTyped/DefinitelyTyped#44946](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44946), so no test results. Spent a bit of time trying things out, but got stuck.
> 
> Could you supply the exact error message? Does this only happen on this feature branch or also on the main branch? In any case, would be great if you could open an issue with the exact error message when trying to compile.
> 

I just tried to recreate this, and could not.  I have made some minor changes to my machine over the last few days, and am thinking one of them must have resolved this.  ( Or it was something wacky from a dependant package, like the recent issue with the mqtt package and its dependencies  ) 

> > Ended up hacking out the current homebridge behaviour where if an advertiser is not declared it defaults to bonjour.
> 
> Yeah, sorry for not clarifying this. Homebridge wasn't updated for this feature yet. This PR is the first step.

Not an issue, just took a bit of effort.

> 
> > After looking at how the Homebridge default advertiser logic works, this may need some further thought as to how we need or want it behave.
> 
> Are you referring to the current **homebridge** default advertiser selection? This always defaults to the `bonjour` package and isn't updated to allow `hap-nodejs` to run its own default advertiser selection.

This is similar to Oznu's comment about being thoughtful about the implementation.

PR #927, opened 2022-01-19
Use bonjour advertiser as default again, warn when Avahi is used on an unsupported platform.

Supereg:

For the beta period of 0.10.0 we changed how we select the default advertiser, moving from `bonjour` to `avahi` as a default if we our heuristic detects it being available on the platform.
This heuristic might not be perfect and might result in non advertised HAP service (see https://github.com/homebridge/homebridge/pull/3062#issuecomment-1014326006).

## :bulb: Proposed solution

To avoid any breaking changes for existing users, this PR revers the default advertiser back to the `bonjour` package.

## :gear: Release Notes

* Revert to `bonjour` as the default advertiser.

When selecting the `avahi` advertiser, this PR will now additionally run the availability check, and revert to `bonjour` if we find to be running on a non compatible platform. Additionally, we print an error message informing the user about it.

### Testing

--

### Reviewer Nudging

--

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -1220,10 +1220,14 @@ export class Accessory extends EventEmitter {
     // create our Advertiser which broadcasts our presence over mdns
     const parsed = Accessory.parseBindOption(info);
 
-    const defaultAdvertiser = info.advertiser
-      ?? (await AvahiAdvertiser.isAvailable() ? MDNSAdvertiser.AVAHI : MDNSAdvertiser.BONJOUR);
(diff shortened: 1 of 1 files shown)


codyc1515:
What is the reasoning for non using ciao?

NorthernMan54:
Pls read the commentary from @adriancable in regards to advertiser issues, #918

Supereg:
> What is the reasoning for non using ciao?

You may also refer to the [Notice on mdns responders](https://github.com/homebridge/ciao#notice-on-native-mdns-responders) in the ciao repo for some context.

Issue #937, opened 2022-04-12
Minor warning message bug in setProps()


thoukydides:
### Analysis

The warning message for `props.maxValue` being below the minimum for the `format` is incorrectly using `numericUpperBound` instead of `numericLowerBound` on [src/lib/Characteristic.ts line 1137](https://github.com/homebridge/HAP-NodeJS/blob/f6629643e481032f2b166ca797315eacbea232ec/src/lib/Characteristic.ts#L1137):
```
            "Characteristic Property 'maxValue' was set to " + props.maxValue + ", but for numeric format " +
            this.props.format + " minimum possible is " + numericUpperBound(this.props.format),
```

This only affects the warning message. The actual range check is applied correctly.

The equivalent code for `props.minValue` is correct.

### Expected Behavior

Report the minimum allowed value rather than the maximum, when the minimum limit is exceeded.

### Steps To Reproduce

n/a

### Logs

```markdown
n/a (spotted during code inspection)
```

### Configuration

```json
n/a
```

### Environment

n/a

### Process Supervisor

not applicable

NorthernMan54:
@thoukydides tsk for spotting this, could you supply the fix as a pull request ?

PR #938, opened 2022-04-13
Corrected message for MaxValue<numericLowerBound (#937)

thoukydides:

The warning message for `props.maxValue` being below the minimum for the `format` is incorrectly using `numericUpperBound` instead of `numericLowerBound`. This only affects the warning message; The actual range check is applied correctly.

The equivalent code for `props.minValue` is correct.

## :bulb: Proposed solution

Replace `numericUpperBound` by `numericLowerBound` when constructing the warning message.

## :gear: Release Notes

Corrected warning message when an attempt is made to set `props.maxValue` to less than the minimum allowed value for the characteristic's type.

### Testing

There do not appear to be any tests for the content of warning messages.


Supereg (approved):
Thanks for correcting this ☺️

Supereg (approved):

PR #944, opened 2022-04-23
HKSV: Skip updateRecordingActive delegate call if value didn't change

Supereg:

As of now, every write to the `Active` characteristic of the `CameraRecordingManagement` service is plainly forwarded to the `CameraRecordingDelegate`. HomeHub write to this characteristic a lot, even though state isn't changed.

## :bulb: Proposed solution

This PR proposes to include a check to determine if the `active` value changed before calling the delegate. This will remove the burden on developers to do this check themselves.

## :gear: Release Notes

* `CameraRecordingDelegate.updateRecordingActive()` is now only ever called if the value actually changed. This is considered a breaking change. It is continued to be called on every startup if recording is enabled, and it will continue to be called on a factory reset (e.g. when all pairings are removed) independent of the current state.

This PR includes a breaking change and is planned for the next major/minor release.

### Testing

--

### Reviewer Nudging

CC @hjdhjd


src/lib/camera/RecordingManagement.ts
```diff
@@ -436,7 +441,14 @@ export class RecordingManagement {
       .setProps({ adminOnlyAccess: [Access.WRITE] });
 
     this.recordingManagementService.getCharacteristic(Characteristic.Active)
-      .onSet(value => this.delegate.updateRecordingActive(!!value))
+      .onSet(value => {
+        if (!!value === this.recordingActive) {
+          return; // skip delegate call if state didn't change!
```
Supereg, line 446: We might want to do a similar treatment with `updateRecordingConfiguration`. Such that developers are not required to check differences in configuration. We can do this efficiently in HAP-NodeJS.

Supereg: Resolved in 8f04fa0

TecnologiaMundia:
<html>
<body>


"@context": "http://schema.org",
--
  | "@type": "Organization",
  | "contactPoint": [{
  | "@type": "ContactPoint",
  | "telephone": "+1874912",
  | "contactType": "customer service",
  | "email": "vozdelcliente@bpd.com.do"
  | }],
  | "sameAs": ["http://www.facebook.com/bancopopulardominicano",
  | "http://www.twitter.com/!#/popularenlinea",
  | "http://www.youtube.com/user/Popularenlinea", "http://www.instagram.com/popularenlinea",
  | "https://es.wikipedia.org/wiki/Banco_Popular_Dominicano"
  | ],
  | "potentialAction": {
  | "@type": "SearchAction",
  | "target": "https://www.popularenlinea.com/personas/Paginas/Search.aspx#k={search_term_string}",
  | "query-input": "required name=search_term_string"


</body>
</html>

PR #946, opened 2022-04-25
fixes #945

### LINKED ISSUE #945: Info: Affacted by network family change in Node.js 18 : "IPv4" is now 4 ??
### Analysis

mainly an info because I found IPv4 as String in the code... 

See: https://github.com/nodejs/node/pull/41431#pullrequestreview-929717323

### Expected Behavior

Considers when working on Node.js 18 compatibility

### Steps To Reproduce

n/a

### Logs

```markdown
n/a
```

### Configuration

```json
n/a
```

### Environment

Node.js 18

### Process Supervisor

not applicable

### Additional Context

_No response_

mieszkomularczyk:

This code does not work with nodejs 18, since the one field type in the objects returend by os.networkInterfaces() changed from string to number. 'family' field is either 4 or 6, previously it was a string IPv4 or IPv6

## :bulb: Proposed solution

add extra check in the conditions so both old and new nodeJS work

### DIFF
--- a/src/lib/util/net-utils.ts
+++ b/src/lib/util/net-utils.ts
@@ -13,11 +13,11 @@ export function findLoopbackAddress(): string {
       }
 
       internal = true;
-      if (info.family === "IPv4" || info.family === 4) { //nodejs18 uses numbers 4 or 6
+      if (info.family === "IPv4" || info.family == 4) { //nodejs18 uses numbers 4 or 6
         if (!ipv4) {
           ipv4 = info.address;
         }
-      } else if (info.family === "IPv6" || info.family === 6) {
+      } else if (info.family === "IPv6" || info.family == 6) {
         if (info.scopeid) {
           if (!ipv6LinkLocal) {
             ipv6LinkLocal = info.address + "%" + name; // ipv6 link local addresses are only valid with a scope
--- a/src/lib/util/net-utils.ts
+++ b/src/lib/util/net-utils.ts
@@ -13,11 +13,12 @@ export function findLoopbackAddress(): string {
       }
 
       internal = true;
-      if (info.family === "IPv4" || info.family == 4) { //nodejs18 uses numbers 4 or 6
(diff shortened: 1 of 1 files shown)


Apollon77:
Sere there not also other places (in logging too) where this was used as string?

Supereg:
I'm not sure on the state of development of node v18. Is this a breaking change in node v18 which is here to stay? The linked comment on https://github.com/nodejs/node/pull/41431#pullrequestreview-929717323 sounded like it may be revised such that it doesn't affect current packages(?).

Apollon77:
I read that one issue like "we want to do the change and should asjut affected libraries" ... eklse I would have expect3d the PR not to be accepted ... or?!

oznu:
@Supereg it's probably worth doing sooner rather than later. There may be other compatibility issues in Node.js 18 we won't find until this one is patched.

oznu:
Maybe add some `@ts-expect-error` notices until we can get updated versions of `@types/node`? Using `@ts-expect-error` will ensure the exception is removed when it is no longer required.

```ts
      internal = true;
      // @ts-expect-error Nodejs 18+ uses the numbers 4 or 6 instead of the strings "IPv4" or "IPv6"
      if (info.family === "IPv4" || info.family === 4) {
        if (!ipv4) {
          ipv4 = info.address;
        }
      // @ts-expect-error Nodejs 18+ uses the numbers 4 or 6 instead of the strings "IPv4" or "IPv6"
      } else if (info.family === "IPv6" || info.family === 6) {
        if (info.scopeid) {
          if (!ipv6LinkLocal) {
            ipv6LinkLocal = info.address + "%" + name; // ipv6 link local addresses are only valid with a scope
          }
        } else if (!ipv6) {
          ipv6 = info.address;
        }
      }
```

mieszkomularczyk:
ok, it does not build anyway (kudos for the one who would explain why:) , therefore I close it until 18 reaches LTS

PR #947, opened 2022-04-26
Fix #945 - Node.js v18 os.networkInterfaces()[].family number/string comparison

### LINKED ISSUE #945: Info: Affacted by network family change in Node.js 18 : "IPv4" is now 4 ??

oznu:

https://github.com/nodejs/node/issues/42787
https://github.com/nodejs/node/pull/41431#discussion_r841277378

## :bulb: Proposed solution

It's worth doing this sooner rather than later. There may be other compatibility issues in Node.js 18 we won't find until this one is patched.

The use of `@ts-expect-error` should alert us to remove the exception when @types/node is updated to 18 (if they decide to keep backwards compatibility with old versions, which they may not, be we need to).

Alternatively we could do something like this, but I feel like it will result in the exception never being removed in the future:

```ts
(info.family === "IPv6" || (info.family as unknown as string) === 6)
```

## :gear: Release Notes

* Fix an issue with network interface family detection when running Node.js 18

### Testing

Tests are still failing on Node v18.0.0 - out of scope for this PR.

### Reviewer Nudging

@Supereg

### DIFF
--- a/src/lib/util/eventedhttp.ts
+++ b/src/lib/util/eventedhttp.ts
@@ -713,7 +713,8 @@ export class HAPConnection extends EventEmitter {
 
     if (ipVersion === "ipv4") {
       for (const info of infos) {
-        if (info.family === "IPv4") {
+        // @ts-expect-error Nodejs 18+ uses the number 4 the string "IPv4"
+        if (info.family === "IPv4" || info.family === 4) {
           return info.address;
         }
       }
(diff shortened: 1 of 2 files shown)


Supereg (approved):
Thanks for addressing this. 👍🏼

Proposing to include this in the upcoming 0.10.2 release. I adopted the base branch accordingly.

PR #950, opened 2022-05-17
Fix `[${this.displayName}]` interpolation for error displayed when Avahi is not available.

oznu:

`[${this.displayName}]` was not being interpolated.

## :bulb: Proposed solution

Use backtick formatting for the entire line.


Supereg (dismissed):
Thanks for catching this.

Supereg (approved):

Issue #955, opened 2022-05-28
Critical floating point error on Characteristic Set


OrigamiDream:
### Analysis

In characteristics like `RotationSpeed` and `Brightness`, we can set the value ranges and the minimum step size.
For instance, an Air Purifier have 6 levels of fan speed, we can set value ranges of RotationSpeed min:0, max:100, and step:16.666... as first option, or min:0, max:6, step:1 as second option.

However, the second option is not working properly with Siri.
The Siri says the value as is, 1%, 2% and 6% max.
To let the Siri works properly, I have to use the first option.

The problem starts here.
If I set the characteristic value manually, the Homebridge runtime changes the value arbitrarily to fit the `minStep` size.
And this must not to be happened.

Details in `Steps to Reproduce`

### Expected Behavior

Characteristic setup:
```typescript
const STEPS = (100 / 6);

service.getCharacteristic(Characteristic.RotationSpeed)
    .setProps({
        minValue: 0.0,
        maxValue: STEPS * 6,
        minStep: STEPS
    })
    .on(CharacteristicEventTypes.SET, (value, callback) => {
        this.log.debug('RotationSpeed SET value', value);
    });
```

Try to change the characteristic value manually using `service#setCharacteristic` for 0 to 100.

`setCharacteristic(RotationSpeed, STEPS * 1)`, expected: 16.6666.., actual value: 16.6666..
`setCharacteristic(RotationSpeed, STEPS * 2)`, **expected: 33.3333.., actual value: 16.6666..**
`setCharacteristic(RotationSpeed, STEPS * 3)`, **expected: 50.0, actual value: 33.3333..**
`setCharacteristic(RotationSpeed, STEPS * 4)`, expected: 66.6666.., actual value: 66.6666..
`setCharacteristic(RotationSpeed, STEPS * 5)`, **expected: 83.3333.., actual value: 66.6666..**
`setCharacteristic(RotationSpeed, STEPS * 6)`, **expected: 100.0, actual value: 83.3333..**

### Steps To Reproduce

1. Create a sample plugin with `AirPurifier` service and `RotationSpeed` characteristic.
2. Set props — minValue: 0, maxValue: 100, minStep: (100 / 6).
3. Try to change the value manually via `service.setCharacteristic()`.
4. Restart the process if the issue does not happen (because floating point precision value changes every process restart in this example)

### Logs

```markdown
This is mathematical and computational issue, no logs.
```

### Configuration

```json
No specific configuration needed
```

[...]

PR #956, opened 2022-05-28
Fix floating point validation on Characteristic Set

OrigamiDream:

Related:
- #955 

## :bulb: Proposed solution

The provided fp value must be same with the characteristic value.

Previously, this didn't work because of remainder between provided value and minimum step size.
Therefore, I suggest a new stable constant, `epsilon`, which is 1.
This patch fixes the issue by reducing remainder significantly.

The value of the constant is `minimum`.
It would crash plugins if the value is under 1.

## :gear: Release Notes

If your accessory have specific "level" of state, you can now change the ranges to 0~100%.
This makes your Siri reads the level of state properly.

No additional information needed

### Testing

"should validate Formats.FLOAT with precision with minimum steps" have been added in `Characteristic.spec.ts#L1097`

### Reviewer Nudging

You don't need hard mathematical understanding!
You can just test it with `eps = 0` or `eps = 1`.

### DIFF
--- a/src/lib/Characteristic.spec.ts
+++ b/src/lib/Characteristic.spec.ts
@@ -1094,6 +1094,35 @@ describe("Characteristic", () => {
       expect(mock).toBeCalledTimes(0);
     });
 
+    it("should validate Formats.FLOAT with precision with minimum steps", () => {
+      const steps = (100 / 6);
+      const characteristic = createCharacteristicWithProps({
+        format: Formats.FLOAT,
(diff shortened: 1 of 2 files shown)


Supereg (approved):
Thanks for reporting and fixing this issue 🚀

adriancable:
@Supereg - see my comments in #955. Unfortunately I think this was merged prematurely. This PR replaces a logic error in the rounding algorithm with another logic error.

PR #958, opened 2022-06-05
Fix floating point precision

OrigamiDream:

Related:
- #955 
- #956 

## :bulb: Proposed solution

As @adriancable have mentioned in #956, the previous solution may have possible incorrect behaviors.
So I have tried all possible tests which can be happened in 0 and 100.

I moved the rounding logic in front of clamping logic.
Plus, to fix in case the error is smaller than epsilon and greater than 0.5, I have changed `Math.round()` to `Math.floor()`. 
We don't need round the value in server-side, due to Apple Home app will always automatically round the value fit to step size.

### Testing

I have rewritten entire test code related with this issue.
The `minStep` in the final test only uses (100/6), of course I have tried other combinations like 3, 7, 9.
All cases have passed, including approaches which @adriancable have mentioned before.

The test takes up to 3 minutes per step, this is the reason why I have decided to use only 100/6.

### Reviewer Nudging

I would appreciate if you have better idea.

### DIFF
--- a/src/lib/Characteristic.spec.ts
+++ b/src/lib/Characteristic.spec.ts
@@ -1095,32 +1095,28 @@ describe("Characteristic", () => {
     });
 
     it("should validate Formats.FLOAT with precision with minimum steps", () => {
-      const steps = (100 / 6);
-      const characteristic = createCharacteristicWithProps({
-        format: Formats.FLOAT,
-        perms: [Perms.PAIRED_READ, Perms.PAIRED_WRITE],
-        minValue: 0,
-        maxValue: 100,
-        minStep: steps,
-      });
+      const pad = 1;
+      const characteristic = createCharacteristic(Formats.FLOAT);
 
-      characteristic.setValue(steps);
-      expect(characteristic.value).toEqual(steps);
+      // takes up to 3 minutes
+      for(let ceil = 1; ceil <= 100; ceil++) {
+        const steps = (ceil / 6);
+        for(let maxValue = ceil; maxValue > 0; maxValue--) {
+          for(let minValue = 0; minValue < maxValue; minValue++) {
+            characteristic.setProps({
+              minValue: minValue,
+              maxValue: maxValue,
+              minStep: steps,
+            });
+            for(let amp = 0; amp < (ceil / steps) + pad; amp++) {
+              const desiredValue = Math.min(Math.max(steps * amp, minValue), maxValue);
 
-      characteristic.setValue(steps * 2);
-      expect(characteristic.value).toEqual(steps * 2);
-
-      characteristic.setValue(steps * 3);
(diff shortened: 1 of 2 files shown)


Supereg:
@adriancable would you mind reviewing this PR?

OrigamiDream:
The failure is originated from the new rounding algorithm. Im gonna fix this after WWDC 🥳

adriancable:
It’s not sufficient to check that stepValue != null. It could also be undefined or 0. Either will lead to an overflow which is what happens in your test.

OrigamiDream:
We should not handle stepValue which is undefined or 0.
The original input value must not be changed in this case.
Instead, we can just warn to the developers.

And the test failure is not a kind of overflow.
They are just the wrong 'numericMin' handling errors in the old test cases.

adriancable:
I think this looks OK now.

src/lib/Characteristic.spec.ts
```diff
@@ -805,6 +805,7 @@ describe("Characteristic", () => {
       const characteristic = createCharacteristicWithProps({
         format: Formats.FLOAT,
         minStep: 0.001,
+        minValue: 0,
```
Supereg, line 808: Was this change required to make the test work again or what’s the reason behind it?

Supereg (commented):
Looks good to me. There were some other edge cases discussed in the issue form right? Would it make sense to add them to the suite of test cases to avoid any future regressions?

src/lib/Characteristic.spec.ts
OrigamiDream: The `minValue` changes to `-Number.MAX_VALUE` internally, which have many decimal points.
Old rounding algorithm does not care about the decimal points, that the reason why the test case had worked.
Since it was working wrongly, I have updated the test case to fit with the new rounding algorithm.

Old algorithm:
- 1.024 -> 1.024 (because the `minValue` is treated as no decimal points)

New algorithm:
- 1.024 + (decimal points of `-Number.MAX_VALUE`) -> 1.7976931348623157**e+308**

To the issue be resolved, the old test case must be changed.

OrigamiDream:
Possibly one edge case.

Basically all characteristics have `minValue` itself, so there would no failure if `minValue` is valid.
However somehow the plugin developer set the `minValue` as `undefined` ignoring default values, the `minValue` will be changed to `-Number.MAX_VALUE`.
Eventually, the rounding algorithm would work wrongly. (It would feel like unexpected behavior to the plugin developer)

Although this can be happened, it is more like the developer sets non-null properties as null which cannot be detected in compile-time.
Therefore, I have decided to not add warning signs about this. (And too many affected test cases which are spying on `characteristicWarning`)

Supereg (approved):

PR #965, opened 2022-08-27
Add systemd-resolved mDNS advertiser support

elyscape:

Homebridge currently supports [three different mDNS advertisers][mdns], two that run in-process and one that uses Avahi over D-Bus. For a while now, systemd-resolved has supported acting as an mDNS advertiser and resolver. If Homebridge is run on a system that uses systemd-resolved for mDNS, this can cause conflicts similar to homebridge/homebridge#3077.

## :bulb: Proposed solution

systemd-resolved, like Avahi, exposes [a D-Bus interface][dbus] for managing mDNS services. This pull request adds a systemd-resolved mDNS advertiser that makes use of this interface. The code was heavily based on that of the Avahi advertiser.

## :gear: Release Notes

- Added a new mDNS advertiser named `resolved` that uses systemd-resolved via D-Bus.

In order for homebridge to have the permission to use this interface, a `.pkla` file with the following contents will need to be installed to `/etc/polkit-1/localauthority/10-vendor.d/`:
```
[Homebridge RegisterService]
Identity=unix-user:homebridge
Action=org.freedesktop.resolve1.register-service
ResultAny=yes

[Homebridge UnregisterService]
Identity=unix-user:homebridge
Action=org.freedesktop.resolve1.unregister-service
ResultAny=yes
```

### Testing

Seems to work for me!

See: homebridge/homebridge#3224

[mdns]: https://github.com/homebridge/homebridge/wiki/mDNS-Options
[dbus]: https://www.freedesktop.org/software/systemd/man/org.freedesktop.resolve1.html

### DIFF
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -264,6 +264,46 @@ export class BonjourHAPAdvertiser extends EventEmitter implements Advertiser {
 
 }
 
+function messageBusConnectionResult(bus: MessageBus): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const errorHandler = (error: Error) => {
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      bus.connection.removeListener("connect", connectHandler);
+      reject(error);
+    };
+    const connectHandler = () => {
+      bus.connection.removeListener("error", errorHandler);
+      resolve();
+    };
+
+    bus.connection.once("connect", connectHandler);
+    bus.connection.once("error", errorHandler);
+  });
+}
+
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function dbusInvoke(bus: MessageBus,
+                    destination: string,
+                    path: string,
+                    interface: string,
+                    member: string,
+                    others?: any): Promise<Any> {
+  return new Promise((resolve, reject) => {
+    const command = {destination, path, interface, member, ...(others || {})};
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    bus.invoke(command, (err: any, result: any) => {
+      if (err) {
+        reject(new Error(`dbusInvoke error: ${JSON.stringify(err)}`));
+      } else {
+        resolve(result);
+      }
+    });
+
+  });
+}
+
 /**
  * Advertiser based on the Avahi D-Bus library.
  * For (very crappy) docs on the interface, see the XML files at: https://github.com/lathiat/avahi/tree/master/avahi-daemon.
@@ -370,7 +410,7 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
 
     try {
       try {
-        await this.messageBusConnectionResult(bus);
+        await messageBusConnectionResult(bus);
       } catch (error) {
         debug("Avahi/DBus classified unavailable due to missing dbus interface!");
         return false;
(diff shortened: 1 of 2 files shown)


elyscape:
Dealt with conflicts from master and ESLint warning. This should be ready whenever.

src/lib/Advertiser.ts
```diff
@@ -422,3 +422,171 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
...
+
+  public async updateAdvertisement(silent?: boolean): Promise<void> {
+    if (!this.bus) {
+      throw new Error("Tried to update systemd-resolved advertisement on a destroyed advertiser!");
+    }
+
+    debug("Updating txt record (txt: %o, silent: %d)", CiaoAdvertiser.createTxt(this.accessoryInfo, this.setupHash), silent);
+
+    await this.stopAdvertising();
+    await this.startAdvertising();
```
Supereg, line 493: Is there no better way to update the txt record with `systems-resolved`? Readvertising the whole service is a pretty heavy operation and is actively discouraged for updates!

src/lib/Advertiser.ts
```diff
@@ -422,3 +422,171 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
...
+    }
+
+    if (this.path) {
+      try {
+        await ResolvedAdvertiser.resolvedInvoke(this.bus, "UnregisterService", {
+          body: [this.path],
+          signature: "o",
+        });
+      } catch (error) {
+        // Typically, this fails if e.g. systemd-resolved service was stopped in the meantime.
```
Supereg, line 508: I'm currently implementing a mechanism to detect restarts of the avahi-daemon for the avahi-based dbus advertiser (see https://github.com/homebridge/HAP-NodeJS/pull/970). Could we pull off something similar with systems-resolved interface?

src/lib/Advertiser.ts
```diff
@@ -422,3 +422,171 @@ export class AvahiAdvertiser extends EventEmitter implements Advertiser {
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      bus.invoke(command, (err: any, result: any) => {
+        if (err) {
+          reject(new Error(`resolvedInvoke error: ${JSON.stringify(err)}`));
+        } else {
+          resolve(result);
+        }
+      });
+    });
+  }
```
Supereg, line 591: This is just a plain copy of the code of the avahi-based advertiser right? Could we try to reuse the existing code a bit more?

Supereg (commented):
Thanks for your contribution 🚀 
Had only a few comments on certain places ☺️ 

You mentioned, that there is some initial setup required to set permissions for homebridge/hap-nodejs. Can we please add this to the in-code documentation so users can easily find that.
Further, can programatically detect, if we are missing the permission for systemd-resolved such that we can issue an according warning (maybe with a link to a Wiki documentation page illustration how to configure permissions)?

src/lib/Advertiser.ts
elyscape: Unfortunately, this is currently the case.

src/lib/Advertiser.ts
elyscape: Possibly. I'd need some guidance on what your PR actually does, and then I'd need to test to see if it worked for systemd-resolved, but in theory it should be doable.

src/lib/Advertiser.ts
elyscape: What would you recommend as the best way to do that?

elyscape:
> Can we please add this to the in-code documentation so users can easily find that.

Sure. Where should I put it?

> Further, can programatically detect, if we are missing the permission for systemd-resolved such that we can issue an according warning (maybe with a link to a Wiki documentation page illustration how to configure permissions)?

If we're missing the appropriate permissions, there are a number of different errors that can occur as a result. Rather than try to figure out what they all are, how do you feel about making the error handler always issue that warning in response to failures?

src/lib/Advertiser.ts
elyscape: I added a comment explaining this.

src/lib/Advertiser.ts
elyscape: I extracted these methods into shared functions.

src/lib/Advertiser.ts
elyscape: I did some digging and unfortunately systemd-resolved's D-Bus interface doesn't have any signals for us to listen onto. As far as I can tell, our best bet to detect restarts is to periodically poll for the existence of the object on the bus and re-register if it's missing.

elyscape:
@Supereg Any thoughts on what else I should do to get this in a mergable state?

elyscape:
Tests should work now, btw.

src/lib/Advertiser.ts
Supereg: > I did some digging and unfortunately systemd-resolved's D-Bus interface doesn't have any signals for us to listen onto.

That's a pity. 

> As far as I can tell, our best bet to detect restarts is to periodically poll for the existence of the object on the bus and re-register if it's missing.

But also this seams like there is not an easy way for us to check this right? The only method that takes the `service_path` as an input is the `UnregisterService` method.
So I think the best is to just document that for now and treat that feature as an experimental feature as for now?

Supereg (approved):

donavanbecker:
@Supereg, I have opened this Pull Request on the UI to add resolved: https://github.com/oznu/homebridge-config-ui-x/pull/1427

Issue #971, opened 2022-09-18
v0.10.3 breaks FLOAT characteristic with minStep of 0.01


ebaauw:
### Analysis

See https://github.com/ebaauw/homebridge-rpi/issues/143.  I'm getting:
> This plugin generated a warning from the characteristic 'Load': characteristic was supplied illegal value: number Infinity exceeded maximum of 1.7976931348623157e+308. See https://homebridge.io/w/JtMGR for more info.

When setting _Load_ to a valid value, like 0.12, 0.2, or 0.07.

The characteristic is defined with type `FLOAT` and `minStep` 0.01.  No `maxValue` nor `minValue`.  I don't recognise `Infinity`, nor the maximum value mentioned in the warning.

After manually reverting to v0.10.2, the warning is gone.

### Expected Behavior

Accept valid values, as v0.10.2.

### Steps To Reproduce

Re-install Homebridge, so it installs hap-nodejs@0.10.3.  Run Homebridge RPi.

### Logs

```markdown
[9/18/2022, 10:15:00 PM] [RPi] pi10: set Frequency from 800MHz to 700MHz
[9/18/2022, 10:15:00 PM] [RPi] pi10: set Load from 0.12 to 0.09
[9/18/2022, 10:15:00 PM] [homebridge-rpi] This plugin generated a warning from the characteristic 'Load': characteristic was supplied illegal value: number Infinity exceeded maximum of 1.7976931348623157e+308. See https://homebridge.io/w/JtMGR for more info.
[9/18/2022, 10:15:00 PM] [homebridge-rpi] Error: 
    at Characteristic.characteristicWarning (/usr/local/lib/node_modules/homebridge/node_modules/hap-nodejs/src/lib/Characteristic.ts:2138:105)
    at Characteristic.validateUserInput (/usr/local/lib/node_modules/homebridge/node_modules/hap-nodejs/src/lib/Characteristic.ts:2054:14)
    at Characteristic.updateValue (/usr/local/lib/node_modules/homebridge/node_modules/hap-nodejs/src/lib/Characteristic.ts:1457:20)
    at CharacteristicDelegate.set value [as value] (/Users/ebaauw/GitHub/homebridge-rpi/node_modules/homebridge-lib/lib/CharacteristicDelegate.js:317:28)
    at Object.set (/Users/ebaauw/GitHub/homebridge-rpi/node_modules/homebridge-lib/lib/ServiceDelegate/index.js:256:50)
[...]

PR #972, opened 2022-09-19
Fixed a regression where floats were rounded incorrectly when no minValue was provided

Supereg:

The release of v0.10.3 introduced fixes to the float rounding mechanism (see #956, #958). It introduced a regression when no `minValue` was provided (see #971).

## :bulb: Proposed solution

This PR introduces a fix for this regression.

## :gear: Release Notes

* Fixed a regression where floats were rounded incorrectly when no `minValue` was provided

This regression slipped through in this review step: https://github.com/homebridge/HAP-NodeJS/pull/958/files#r893353790.

### Testing

A regression test was added, covering this case.

### Reviewer Nudging

--

### DIFF
--- a/src/lib/Characteristic.spec.ts
+++ b/src/lib/Characteristic.spec.ts
@@ -1063,6 +1063,22 @@ describe("Characteristic", () => {
       expect(characteristic.value).toEqual(0.1);
     });
 
+    it("should accept Formats.FLOAT with non-defined min/max value", () => {
+      const characteristic = createCharacteristicWithProps({
+        format: Formats.FLOAT,
+        minStep: 0.01,
+        perms: [Perms.PAIRED_READ, Perms.NOTIFY],
+      }, uuid.generate("051"));
+
+      // @ts-expect-error - spying on private property
+      const mock = jest.spyOn(characteristic, "characteristicWarning");
+
+      mock.mockReset();
+      characteristic.updateValue(0.09);
+      expect(characteristic.value).toEqual(0.09);
+      expect(mock).toBeCalledTimes(0);
+    });
+
     it("should validate Formats.FLOAT with precision", () => {
       const characteristic = new Characteristic.CurrentAmbientLightLevel();
 
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -602,9 +602,9 @@ function extractHAPStatusFromError(error: Error) {
 }
 
 function maxWithUndefined(a?: number, b?: number): number | undefined {
-  if (a === undefined) {
+  if (a == null) {
     return b;
-  } else if (b === undefined) {
+  } else if (b == null) {
     return a;
   } else {
     return Math.max(a, b);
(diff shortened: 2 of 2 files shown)


Supereg:
CC: @OrigamiDream

OrigamiDream:
In my opinion, `minValue` and `maxValue` must be **defined** in characteristic level.

Current state:
```typescript
const numericMin = maxWithUndefined(this.props.minValue, numericLowerBound(this.props.format));
const numericMax = minWithUndefined(this.props.maxValue, numericUpperBound(this.props.format));
```

And new suggestion:
```typescript
/* Needs to find original props */
const originalMin = originalProps.minValue; // if there is no defined default values, code must be failed.
const originalMax = originalProps.maxValue;
const numericMin = Math.max(numericLowerBound(this.props.format), maxWithUndefined(this.props.minValue, originalMin));
const numericMax = Math.min(numericUpperBound(this.props.format), maxWithUndefined(this.props.maxValue, originalMax));
```
This is not only preventing setting values to exceeding their limits, but also falling back to specific reasonable default value when the developer haven't set its specific value.
No other changes needed except finding original properties.

All characteristic have their min and max values in their definitions.
But what if developers ignore those values on specific purpose?
The code would be failed because it is not correct operation.
Therefore, those values must be `required`, not `optional` even custom characteristic.

We have to restrict this to developers, not avoiding faced problem.

ebaauw:
> In my opinion, minValue and maxValue must be defined in characteristic level.

I don’t disagree, but then please issue a warning when these are missing instead.

ebaauw (commented):
Also happy to see a warning when minValue and maxValue are missing, and leave the 0.10.3 logic.

I do remember issues with setting minValue for temperature to -273.15, because the minStep 0.1 would indeed start there, and no longer accept 20.5.

Supereg:
> I do remember issues with setting minValue for temperature to -273.15, because the minStep 0.1 would indeed start there, and no longer accept 20.5.

That's correct behaviour then.

What I'm proposing is the following logic:

1) `minStep` is **always** (by specification) relative to the `minValue`. Meaning you example I cited above is correct to be rejected.
2) When the developer doesn't specify a `minValue`, we want to have a) to allow a value in the whole number space (e.g. choosing the lowest possible value for the data type) but b) want to use `0` as the reference point for `minStep` (this feels the most natural to me and what developers would expect if they set no value restrictions).

Requiring the `minValue` to be set is a regression as a) the change wasn't described in the previous PR b) there is no check if `minValue` is set, no warnings and no documentation about it (we even have fallback code for other things not requiring `minValue` to be set) and c) this change isn't allowed from a SemVer perspective.

Therefore, I would propose to merge this PR as is for a 0.10.4 release. We could open a discussion separately if we should change the behaviour for a future minor or major version increment of HAP-NodeJS.

OrigamiDream:
If this PR doesn't break any logics from 0.10.3 when `minValue` is specified, this is good to be merged for hot fix purpose.
And well, I had mentioned about this problem when my previous PR have been reviewed.
- https://github.com/homebridge/HAP-NodeJS/pull/958#issuecomment-1152231069

Supereg:
> If this PR doesn't break any logics from 0.10.3 when minValue is specified, this is good to be merged for hot fix purpose.

This PR only changes the logic in the case where `minValue` isn't present. So it doesn't alter any behaviour corrected in any of the previous PPs.

ebaauw (approved):

Issue #980, opened 2022-10-31
Doesn't fallback from systemd-resolved advertiser if MulticastDNS is disabled.


Supereg:
### Analysis

For some (many?) configurations systemd-resolved `MulticastDNS` option is disabled by default. See the `MulticastDNS` option: https://linuxreviews.org/Systemd-resolved.

This results in the following error when starting up with `RESOLVED` advertiser enabled in HAP-NodeJS:
```
Could not create mDNS advertisement. The HAP-Server won't be discoverable: Error: dbusInvoke error: ["Support for MulticastDNS is disabled"]
```

Also the original PR #965 added instructions to enable permissions for mdns advertising. This should be explained in the documentation!

### Expected Behavior

This should be checked at startup within the `ResolvedAdvertiser.isAvailable()` function, if `RESOLVED` is actually available on the platform, in order to fall back to a supported advertiser in the erroneous case.

### Steps To Reproduce

1. Setup `/etc/systemd/resolved.conf` to disable the `MulticastDNS` option.
2. Startup HAP-NodeJS instance with advertiser set to `MDNSAdvertiser.RESOLVED`.
3. See the supplied output.

### Logs

```markdown
Could not create mDNS advertisement. The HAP-Server won't be discoverable: Error: dbusInvoke error: ["Support for MulticastDNS is disabled"]
```

### Configuration

```json
-
```

### Environment

- OS: Ubuntu 20.04
- Software: 0.11.0 beta
- Node: v10.19.0
- npm: 6.14.4

### Process Supervisor

not applicable

### Additional Context

CC: @elyscape

elyscape:
Good catch. It should be fairly simple to add a check for this to `isAvailable()`. I'll give it a go tomorrow.

Supereg:
Is there a way to query the systmed dbus interface to check if the MulticastDNS option is enabled? Or is the only way to probe if we are successful in advertising our service?

Also, could you write a short abstract for this Wiki page (https://github.com/homebridge/homebridge/wiki/mDNS-Options) about what steps you need to do to configure the resolved advertiser (e.g. including the link above on how to enable the MulticastDNS option and also to set the permissions so homebridge/hap-nodejs is actually allowed to access it).

Lastly, what error message is thrown if the respective user doesn't have permissions to access the interface? We should definitely also catch that in the `isAvailable` method.

elyscape:
> Is there a way to query the systmed dbus interface to check if the MulticastDNS option is enabled?

[...]

PR #982, opened 2022-11-02
Ensure systemd-resolved has mDNS enabled

elyscape:

Per #980, the systemd-resolved advertiser reports that it's available even if systemd-resolved has mDNS disabled.

## :bulb: Proposed solution

Check that mDNS is enabled in systemd-resolved before reporting that it is available as an advertiser.

## :gear: Release Notes

--

### Testing

--

### Reviewer Nudging

--

### Notes

This also adds a note to the error message when advertisements fail that there may be a permissions issue and the user should check the wiki for more details. The info to be added to the wiki will be posted in #980.

### DIFF
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -279,12 +279,16 @@ function messageBusConnectionResult(bus: MessageBus): Promise<void> {
 }
 
 export class DBusInvokeError extends Error {
-  readonly name: string;
-  readonly message: string;
+  readonly errorName: string;
 
   constructor(errorObject: InvokeError) {
-    super(`dbusInvoke error: ${JSON.stringify(errorObject)}`);
-    this.name = errorObject.name;
+    super();
+
+    Object.setPrototypeOf(this, DBusInvokeError.prototype);
+
+    this.name = "DBusInvokeError";
+
+    this.errorName = errorObject.name;
 
     if (Array.isArray(errorObject.message) && errorObject.message.length === 1) {
       this.message = errorObject.message[0];
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -671,7 +671,7 @@ export class ResolvedAdvertiser extends EventEmitter implements Advertiser {
           },
         );
 
-        if (mdnsStatus[0].type !== "s") {
+        if (mdnsStatus[0][0].type !== "s") {
           throw new Error("Invalid type for MulticastDNS");
         }
 
--- a/src/lib/Advertiser.ts
+++ b/src/lib/Advertiser.ts
@@ -2,7 +2,7 @@
 /// <reference path="../../@types/bonjour-hap.d.ts" />
 import ciao, { CiaoService, MDNSServerOptions, Responder, ServiceEvent, ServiceTxt, ServiceType } from "@homebridge/ciao";
 import { InterfaceName, IPAddress } from "@homebridge/ciao/lib/NetworkManager";
-import dbus, { DBusInterface, MessageBus } from "@homebridge/dbus-native";
+import dbus, { DBusInterface, InvokeError, MessageBus } from "@homebridge/dbus-native";
 import assert from "assert";
 import bonjour, { BonjourHAP, BonjourHAPService, MulticastOptions } from "bonjour-hap";
 import crypto from "crypto";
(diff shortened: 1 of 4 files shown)


elyscape:
cc: @Supereg

src/lib/Advertiser.ts
```diff
@@ -626,21 +631,50 @@ export class ResolvedAdvertiser extends EventEmitter implements Advertiser {
...
+      dbusInterface,
       member,
       others,
     );
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private static resolvedInvoke(bus: MessageBus, member: string, others?: any): Promise<any> {
+    return this.managerInvoke(bus, "org.freedesktop.resolve1.Manager", member, others);
+  }
```
Supereg, line 679: I fell like those two method names are mixed up, if I'm not wrong?

src/lib/Advertiser.ts
```diff
@@ -626,21 +631,50 @@ export class ResolvedAdvertiser extends EventEmitter implements Advertiser {
...
+            body: ["org.freedesktop.resolve1.Manager", "MulticastDNS"],
+            signature: "ss",
+          },
+        );
+
+        if (mdnsStatus[0].type !== "s") {
+          throw new Error("Invalid type for MulticastDNS");
+        }
+
+        if (mdnsStatus[1][0] !== "yes" ) {
```
Supereg, line 649: Looking at the documentation:

`The LLMNR and MulticastDNS properties report whether LLMNR and MulticastDNS are (globally) enabled. Each may be one of "yes", "no", and "resolve". See SetLinkLLMNR() and SetLinkMulticastDNS() above.`

May we even be able to set the MulticastDNS option considering we have the permission for it?

Supereg (commented):
Thanks a lot for dealing with this so fast 🚀 

Looks good to me. Only had two small comments.

src/lib/Advertiser.ts
elyscape: Yeah that's fair. I'll swap them.

src/lib/Advertiser.ts
elyscape: We potentially could but we shouldn't. The most common reason for it to be set to `no` or `resolve` is that some other mDNS responder, like Avahi, is already running on the same machine. If that's the case, I seem to recall the behavior being that setting it to `yes` will succeed but after a brief period of time it will automatically revert to the previous value. Also, we probably don't want to mess with the user's system configuration ourselves.

Supereg: You are completely right. 👍 I didn't think about the implications.

src/lib/Advertiser.ts
```diff
@@ -547,18 +547,23 @@ export class ResolvedAdvertiser extends EventEmitter implements Advertiser {
...
+          "_hap._tcp", // type
+          this.port, // service_port
+          0, // service_priority
+          0, // service_weight
+          [this.createTxt()], // txt_datas
+        ],
+        signature: "sssqqqaa{say}",
+      });
+    } catch (error) {
+      error.message = `Possible permissions issue. Check the wiki for details. ${error.message}`;
```
Supereg, line 564: You may add the link shorthand link to the Wiki page: https://homebridge.io/w/mDNS-Options :)

Supereg:
Just released `@homebridge/dbus-native` [v0.5.0](https://github.com/homebridge/dbus-native/releases/tag/v0.5.0) with the respective change implemented in https://github.com/homebridge/dbus-native/pull/2

src/lib/Advertiser.ts
```diff
@@ -626,21 +645,50 @@ export class ResolvedAdvertiser extends EventEmitter implements Advertiser {
...
+          bus,
+          "org.freedesktop.DBus.Properties",
+          "Get",
+          {
+            body: ["org.freedesktop.resolve1.Manager", "MulticastDNS"],
+            signature: "ss",
+          },
+        );
+
+        if (mdnsStatus[0].type !== "s") {
```
Supereg, line 659: When I execute the advertiser with a default configuration `mdnsStatus` has the following value for me:
`[ [ { type: 's', child: [] } ], [ 'no' ] ]``. So wouldn't the check need to be corrected like the following:?

```suggestion
        if (mdnsStatus[0][0].type !== "s") {
```

elyscape: Probably, yeah. I'll have this PR revised sometime today.

Supereg: I think the PR is already in a good state. We could maybe improve on presenting/rendering the error object (currently it's just json serialized). The only big thing missing is the wiki post.
Other than that I'm very happy with including the PR into the beta branch 🎉 

Thanks a lot for the work you put into 🚀

elyscape:
Okay, this should be good to go now.

Supereg (approved):
Thanks a lot 🚀

PR #1000, opened 2023-03-31
Please Merge New Types for WaterQuality

waltmoorhouse:

Water Quality Sensors are not supported

## :bulb: Proposed solution

Add a Service and Characteristics to allow Water Quality Sensors to be supported.

## :gear: Release Notes

New Types added. No breaking changes.

### Reviewer Nudging

Create a new WaterQualityService and add the new Characteristics to it.

### DIFF
--- a/src/lib/Characteristic.ts
+++ b/src/lib/Characteristic.ts
@@ -853,6 +853,10 @@ export class Characteristic extends EventEmitter {
   /**
    * @group Characteristic Definitions
    */
+  public static Conductivity: typeof Conductivity;
+  /**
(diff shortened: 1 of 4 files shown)


Supereg:
When was this service introduced? Are they documented anywhere?
The files the PR changes are auto generated normally. The sources from which we auto generate those files don't include the updated service.
Otherwise, the generator tool also includes the ability to manually define these inputs.

donavanbecker:
going to close this PR as there hasn't been any movement for over a year. If you would like to provide documentation as @Supereg requested then we can revisit this.

Issue #993, opened 2022-12-14
v0.11.0 fails on Synology


ebaauw:
### Analysis

Homebridge v1.6.0, or rather HAP-NodeJS v0.11.0 doesn't work on my Synlogy.  I'm getting an error `Could not create mDNS advertisement. The HAP-Server won't be discoverable: Error: No such interface found`
Haven't changed anything on my config; reverting to Homebridge v1.5.9 / HAP-NodeJS v0.10.4 solves the issue. 
Using Avahi as advertiser.
I'm running a DS920+ and a DS918+, both on DSM 7.1.1.  Both use link aggregation; the interface name is ovs_bond0.  The Homebridge UI does find this interface, and displays the network activity graph.

### Expected Behavior

HAP-NodeJS should advertise itself.

### Steps To Reproduce

Upgrade Homebridge to v1.6.0 on Synology with `avahi` as advertiser.

### Logs

```markdown
[12/14/2022, 10:39:20 PM] [HB Supervisor] Restarting Homebridge...
[12/14/2022, 10:39:20 PM] [HB Supervisor] Starting Homebridge with extra flags: -P /volume1/homebridge/node_modules -D --strict-plugin-resolution
[12/14/2022, 10:39:20 PM] [HB Supervisor] Starting Homebridge with custom env: {"DEBUG":"HAP-NodeJS:Advertiser","NODE_OPTIONS":"--trace-warnings"}
[12/14/2022, 10:39:20 PM] [HB Supervisor] Started Homebridge v1.6.0 with PID: 26647
2022-12-14T21:39:21.212Z HAP-NodeJS:Advertiser Initializing HAP-NodeJS v0.11.0 ...
[12/14/2022, 10:39:21 PM] Loaded config.json with 0 accessories and 1 platforms.
[12/14/2022, 10:39:21 PM] Loaded 0 cached accessories from cachedAccessories.
[12/14/2022, 10:39:21 PM] ---
[12/14/2022, 10:39:21 PM] No plugins found.
[12/14/2022, 10:39:21 PM] Loading 1 platforms...
[12/14/2022, 10:39:21 PM] Publishing bridge accessory (name: ds2, publishInfo: {
  username: 'CC:22:3D:E3:C7:3D',
  port: 51826,
  pincode: '***-**-***',
  category: 2,
  bind: undefined,
  mdns: undefined,
  addIdentifyingMaterial: true,
  advertiser: 'avahi'
}).
Setup Payload:
[...]

PR #970, opened 2022-09-17
Register listener on the dbus interface to detect Avahi deamon restarts

Supereg:
## :recycle: Current situation

As of now we don't detect restarts of Avahi Deamon and therefore remain without an advertising in this situation as explained in #923.

## :bulb: Proposed solution

This PR fixes this by subscribing on the dbus `StateChanged` listener to detect avahi deamon restarts.

## :gear: Release Notes

* Properly recreate advertisements on restarts of avahi when using the avahi/dbus based advertiser.

### Testing

--

### Reviewer Nudging

--

PR #1003, opened 2023-04-30
Workaround avahi issues on synology platform

bauer-andreas:

As described in #993 there are issues with avahi advertiser feature introduced in #970 where we detect avahi-daemon restarts.
This feature fails on synology platforms and causes the whole advertiser to fail.

## :bulb: Proposed solution

This PR introduces a workaround to be able to start the advertiser even when the restart listener is not in place.

## :gear: Release Notes

* Introduced a workaround for issues with avahi advertiser on Synology

### Testing

--

### Reviewer Nudging

--

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -1430,7 +1430,12 @@ export class Accessory extends EventEmitter {
 
     this._advertiser!.startAdvertising()
       .then(() => this.emit(AccessoryEventTypes.ADVERTISED))
-      .catch(reason => console.error("Could not create mDNS advertisement. The HAP-Server won't be discoverable: " + reason));
+      .catch(reason => {
+        console.error("Could not create mDNS advertisement. The HAP-Server won't be discoverable: " + reason);
+        if (reason.stack) {
+          debug("Detailed error: " + reason.stack);
+        }
+      });
 
     this.emit(AccessoryEventTypes.LISTENING, port, hostname);
   }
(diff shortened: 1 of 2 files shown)


ebaauw:
On my DiskStation running DSM 7.1.1-42962 Update 5, using `avahi` as advertiser still gives an error on startup:
```                  
[4/30/2023, 9:33:19 PM] Homebridge v1.6.1 (HAP v0.11.1) (ds2) is running on port 51826.
Failed to create listener for avahi-daemon server state. The system will not be notified about restarts of avahi-daemon and will therefore stay undiscoverable in those instances. Error message: Error: No such interface found
```
But I do see the mDNS  advertisement in _Discovery_.

For reference, v1.6.0 gives (on my 
```
[4/30/2023, 9:39:59 PM] Homebridge v1.6.0 (HAP v0.11.0) (ds1) is running on port 51826.
Could not create mDNS advertisement. The HAP-Server won't be discoverable: Error: No such interface found
```
And results in no mDSN advertisement.

Both versions issue no errors and show mDNS advertisements when using `ciao` instead.

Supereg:
That is expected when you look at the source code changes.
In the end it is a **workaround** as nobody could identify the real issue behind us not being able to acquire the dBus interface on synology.

I think just printing a warning is the best case for now.

Supereg:
1.6.1 basically restore the functionality pre 1.6.0 (not being able to detect restarts).

Rikezzz:
This update also update node.js?
how to update it node

PR #1010, opened 2023-06-29
Fix naming for Characteristic.ProgramMode

n0rt0nthec4t:
removes extra '_' on end of value Characteristic.ProgramMode.PROGRAM_SCHEDULED_MANUAL_MODE_

Seems extra '_' on end of define was unintentional

## :bulb: Proposed solution

Changes naming Characteristic.ProgramMode.PROGRAM_SCHEDULED_MANUAL_MODE_ to Characteristic.ProgramMode.PROGRAM_SCHEDULED_MANUAL_MODE

## :gear: Release Notes

This maybe a breaking change as removes define for "PROGRAM_SCHEDULED_MANUAL_MODE_". Developers will need to updated any used of Characteristic.ProgramMode.PROGRAM_SCHEDULED_MANUAL_MODE_ to Characteristic.ProgramMode.PROGRAM_SCHEDULED_MANUAL_MODE


Supereg:
I'm not sure if this change is worth deliberately breaking plugins. I would either just leave it as is, or introduce the corrected `PROGRAM_SCHEDULED_MANUAL_MODE` while deprecating the old `PROGRAM_SCHEDULED_MANUAL_MODE_` and have a grace period before completely removing it.

n0rt0nthec4t:
@Supereg fair call.. I think we should have it fixed from a naming consistency view.. I’ll update the commit with the old and new and mark the old as deprecated

n0rt0nthec4t (commented):
Updated to add note PROGRAM_SCHEDULE_MANUAL_MODE_ is deprecated and will be removed

bauer-andreas (approved):
Thanks 🚀

src/lib/definitions/CharacteristicDefinitions.ts
bauer-andreas, line 1: This file is currently getting auto generated. Is there a way of editing the generator file so this doesn’t get overwritten on next generation?

donavanbecker: Ping @n0rt0nthec4t

n0rt0nthec4t: Well... I guess have to work out how they are auto generated.. It can take another year

n0rt0nthec4t: Ok. So in the HAP FramWork plist, this is defined as "Program scheduled (Manual Mode)". So the really comes done to how we handle namings when enclosed in brackets/braces etc. We can correct the auto generation code to account for this?

Shaquu:
Probably not in scope, but I noticed some warning is those example Accessories related to special chars in names. Probably related to recent PR in beta?

src/lib/definitions/CharacteristicDefinitions.ts
bauer-andreas: This file is generated in [`generate-definitions.ts`](https://github.com/homebridge/HAP-NodeJS/blob/853171f84fccf53e56c355e6dd8a808010ebe9ba/src/lib/definitions/generate-definitions.ts) and there are some possibilities for overriding how definitions are handled in [`generator-configuration.ts`](https://github.com/homebridge/HAP-NodeJS/blob/853171f84fccf53e56c355e6dd8a808010ebe9ba/src/lib/definitions/generator-configuration.ts).

Correcting the generation code sounds great. However, we probably want to manually patch the old value as a deprecated constant, such that we do not break backwards compatibility.

n0rt0nthec4t: Thanks @Supereg I'll have a look over and see what we can do for a PR. Seems might take some more work due to structure and filechanges since those were written

donavanbecker:
> Probably not in scope, but I noticed some warning is those example Accessories related to special chars in names. Probably related to recent PR in beta?

yes but should be resolved.

src/lib/definitions/CharacteristicDefinitions.ts
donavanbecker: @n0rt0nthec4t, we are close to release v1.0.0. Are you wanting this included?

n0rt0nthec4t: I wont get todo anything anytime soon. Perhaps we'll close this PR and I'll do a new on at some point to fix the issues from the root cause

donavanbecker: I am going to close the pr as we will resolve this in v1.0.0 and update auto generated docs soon.

PR #1023, opened 2024-02-11
add avahi file advertiser

jneubrand:

- the docker container requires host networking because homebridge wants to advertise its own service

## :bulb: Proposed solution

- make it possible to publish zeroconf information via an avahi xml file, so that it can be mounted, example:
  ```yml
  volumes:
  - /etc/avahi/services/myhub-0000.service:/homebridge/myhub-0000.service
  ```
  (of course, this also requires changes in `config.json` as well as [`jneubrand/homebridge#94b2f53`](https://github.com/jneubrand/homebridge/commit/94b2f538657f064bb8dfd35f67c1679e96cd8dee))
  once this is done, it's no longer necessary to expose host networking when using the docker homebridge image. (other relevant ports need to be exposed explicitly nonetheless)

## :gear: Release Notes

- add avahi-file advertiser

- i've been using this patch for a few years and it works
- should probably allow users to configure the path the `.service` file is written to

### Testing

- none implemented, just want to submit a proof-of-concept pr to check if anyone else wants this :)

### Reviewer Nudging

this isn't ready to merge yet

### DIFF
--- a/src/lib/Accessory.ts
+++ b/src/lib/Accessory.ts
@@ -28,7 +28,7 @@ import {
   VoidCallback,
   WithUUID,
 } from "../types";
-import { Advertiser, AdvertiserEvent, AvahiAdvertiser, BonjourHAPAdvertiser, CiaoAdvertiser, ResolvedAdvertiser } from "./Advertiser";
+import { Advertiser, AdvertiserEvent, AvahiAdvertiser, BonjourHAPAdvertiser, CiaoAdvertiser, ResolvedAdvertiser, AvahiFileAdvertiser } from "./Advertiser";
 // noinspection JSDeprecatedSymbols
 import { LegacyCameraSource, LegacyCameraSourceAdapter, StreamController } from "./camera";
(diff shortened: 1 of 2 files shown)


jneubrand:
Dear maintainers: do you have opinions on how the path for `.service` files should be configured/passed in? I saw `createWithOptions` for `BonjourHAPAdvertiser` as well as the `responderOptions` parameter for `CiaoAdvertiser`, but those never seem to be called from `homebridge/homebridge` itself.

bauer-andreas:
Hey,
Thank you for this contribution. This looks great and very promising for more isolated docker-based installations of Homebridge.
It will probably take some more time until we can have a more closer look at this. We hope to be back around end of March 👍

NorthernMan54:
@jneubrand the GitHub repo docker-homebridge is where the docker image is built from, including the scripts etc. Would it make sense to open this there?

PS - I’m out of the country till mid March so can’t look at this till then.

PR #1030, opened 2024-05-01
Updated dependencies (`axios` and `commander`)

bwp91:
Part of the gradual process of 'major version' dependency updates:

- axios from v0 to v1
- commander from v6 to v12


src/lib/util/eventedhttp.spec.ts
```diff
@@ -30,7 +30,7 @@ describe("eventedhttp", () => {
     await awaitEventOnce(server, EventedHTTPServerEvent.LISTENING);
 
     const address = server.address();
-    baseUrl = `http://${address.address}:${address.port}`;
+    baseUrl = `http://0.0.0.0:${address.port}`;
```
bwp91, line 33: Running this test locally, it returns `address.address` as `::` which makes the baseUrl `http://:::55555` (port randomly chosen here).

The updated version of axios rejects such an address as `Invalid URL`.

I wasn't totally sure with updating these tests to use `0.0.0.0` instead but wasn't sure how else to make the tests pass 🤔

bauer-andreas: I think :: would need to be formatted as [::]:55555. otherwise using 0.0.0.0 should work. Or is it important that we test ipv6 behavior here?

bwp91: Other tests seem to use localhost in fact, unless this is totally separate?

https://github.com/homebridge/HAP-NodeJS/blob/19f80abf3ad9131154c33167cf17c880bf2e46ed/src/lib/HAPServer.spec.ts#L107

Using `[::]` in the base url in the eventedhttp spec file results in:

```
Summary of all failing tests
 FAIL  src/lib/util/eventedhttp.spec.ts
  ● eventedhttp › simple http request

    getaddrinfo ENOTFOUND [::]

```

bauer-andreas: > Using [::] in the base url in the eventedhttp spec file results in:

Hm, interesting. If it works with 0.0.0.0 I'm fine with that 👍

bauer-andreas (approved):

PR #1034, opened 2024-05-19
Updated dependencies (`typescript`)

bwp91:
Updated `typescript` to the latest version, and the `@typescript-eslint/` dependencies too.

---

### Changed

- Updated dependencies (`typescript`)

### DIFF
--- a/src/internal-types.ts
+++ b/src/internal-types.ts
@@ -33,19 +33,23 @@ export function consideredTrue(input: string | null): boolean {
 export const enum TLVValues {
   // noinspection JSUnusedGlobalSymbols
   REQUEST_TYPE = 0x00,
+  // eslint-disable-next-line @typescript-eslint/no-duplicate-enum-values
   METHOD = 0x00, // (match the terminology of the spec sheet but keep backwards compatibility with entry above)
   USERNAME = 0x01,
+  // eslint-disable-next-line @typescript-eslint/no-duplicate-enum-values
   IDENTIFIER = 0x01,
   SALT = 0x02,
   PUBLIC_KEY = 0x03,
   PASSWORD_PROOF = 0x04,
   ENCRYPTED_DATA = 0x05,
   SEQUENCE_NUM = 0x06,
+  // eslint-disable-next-line @typescript-eslint/no-duplicate-enum-values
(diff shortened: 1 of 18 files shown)


donavanbecker (dismissed):

src/internal-types.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-duplicate-enum-values */
```
bauer-andreas: Can we use `esline-disable-next` and attach them to the specific line that triggers the check, if it makes sense here  (e.g., like here https://github.com/homebridge/HAP-NodeJS/blob/110a6924d6c595efb562032a005eca887a7fd961/src/types/dbus-native.d.ts#L10).
This allows us to have a minimal surface where rules are disabled (e.g., and avoid adding non-compliant code in the future). It also helps to clarify which lines actually triggers a warning.

This applies to a lot of other places in this PR as well.

bauer-andreas (dismissed):
Thank you for your work 🚀 

I just had a small suggestion on how we could `eslint-disable-next` comments to have a better overview of what triggered the warning and avoid disabling "too much". 👍

src/internal-types.ts
bwp91: Hi @Supereg yep of course have just updated the PR.

I don't mind either way having the eslint disable rules more specific to the lines they ignore, my initial preference was just to not split the class definition line from the docs above it

https://github.com/homebridge/HAP-NodeJS/blob/95735a759c7f0ba240145a383830bffe4078a406/src/lib/Service.ts#L176-L182

donavanbecker (approved):

src/internal-types.ts
bauer-andreas: There is also `// eslint-disable-line ...` if you want to put the comment into the same line as the warning.

The documentation comment should still attach to the type or is it now dangling and breaks docs generation?

bauer-andreas (approved):

donavanbecker (approved):

src/internal-types.ts
bwp91: the docs are still generating fine

PR #1048, opened 2024-06-25
Bridged core and core cleanup

Shaquu:

*Describe the current situation. Explain current problems, if there are any. Be as descriptive as possible (e.g., including examples or code snippets).*

1. Current code holds incorrect deprecation date, 2022 seems like a little outdated
2. There are some typos
3. There are unhandled promises :)
4. Forbidden char in Sprinkler example!

## :bulb: Proposed solution

*Describe the proposed solution and changes. How does it affect the project? How does it affect the internal structure (e.g., refactorings)?*

1. Current code holds incorrect deprecation date, 2022 seems like a little outdated
2. I did some clean-up regarding deprecation warning
3. I did handled promises!
4. Removed forbidden chars

### DIFF
--- a/src/BridgedCore.ts
+++ b/src/BridgedCore.ts
@@ -6,7 +6,7 @@ import { Accessory, AccessoryEventTypes, AccessoryLoader, Bridge, Categories, HA
 
 console.log(`HAP-NodeJS v${HAPLibraryVersion()} starting...`);
 
-console.warn("DEPRECATION NOTICE: The use of Core and BridgeCore is deprecated and is scheduled to be removed in 2024.");
+console.warn("DEPRECATION NOTICE: The use of Core and BridgeCore is deprecated and is scheduled to be removed in a future version.");
 console.warn("For more information and some guidance on how to migrate, have a look at https://github.com/homebridge/HAP-NodeJS/wiki/Deprecation-of-Core-and-BridgeCore");
 
 // Initialize our storage system
--- a/src/BridgedCore.ts
+++ b/src/BridgedCore.ts
@@ -6,8 +6,8 @@ import { Accessory, AccessoryEventTypes, AccessoryLoader, Bridge, Categories, HA
 
 console.log(`HAP-NodeJS v${HAPLibraryVersion()} starting...`);
 
-console.warn("DEPRECATION NOTICE: The use of Core and BridgeCore are deprecated and are scheduled to be remove in October 2020. " +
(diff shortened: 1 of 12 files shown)


src/BridgedCore.ts
```diff
@@ -6,8 +6,8 @@ import { Accessory, AccessoryEventTypes, AccessoryLoader, Bridge, Categories, HA
 
 console.log(`HAP-NodeJS v${HAPLibraryVersion()} starting...`);
 
-console.warn("DEPRECATION NOTICE: The use of Core and BridgeCore are deprecated and are scheduled to be remove in October 2020. " +
-  "For more information and some guidance on how to migrate, have a look at https://github.com/homebridge/HAP-NodeJS/wiki/Deprecation-of-Core-and-BridgeCore");
+console.warn("DEPRECATION NOTICE: The use of Core and BridgeCore is deprecated and is scheduled to be removed in 2024.");
```
bwp91: i think to keep this generic remote the year and just put “in a future version”

src/BridgedCore.ts
```diff
@@ -42,9 +42,7 @@ const signals = { "SIGINT": 2, "SIGTERM": 15 } as Record<string, number>;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 Object.keys(signals).forEach((signal: any) => {
   process.on(signal, () => {
-    bridge.unpublish();
```
bwp91, line 45: i’m on my phone so can’t properly highlight but was just wondering what this change does

bwp91 (commented):
couple of comments.

i will always veer towards asking if any of the changes you have made could potentially breaking in any way

src/BridgedCore.ts
Shaquu: Will do

src/BridgedCore.ts
Shaquu: So, `unpublish` is async. So AFAIK, it could happen that before it is done doing its job, we would kill the process. Potentially leaving some spills in a system (files? mdns records? Whatever).

In the new version, we will kill process only when `unpublish` is done.

Shaquu:
> couple of comments.
> 
> i will always veer towards asking if any of the changes you have made could potentially breaking in any way

One potential break. I removed `💦` from Accessory and Service.
It was used in `AccessoryInformation` Characteristic.
But from what I see it was not used to generate UUID.

donavanbecker (dismissed):

donavanbecker (approved):

Shaquu:
Ready to be merged?

PR #1049, opened 2024-06-27
Fix ConfiguredName

donavanbecker:
From what I can see `ConfiguredName` should be setup like this.

Also added a log to indicate if the name is empty completely.

### DIFF
--- a/src/lib/util/checkName.ts
+++ b/src/lib/util/checkName.ts
@@ -1,3 +1,5 @@
+import { CharacteristicValue, Nullable } from "../../types";
+
 /**
  * Checks that supplied field meets Apple HomeKit naming rules
  * https://developer.apple.com/design/human-interface-guidelines/homekit#Help-people-choose-useful-names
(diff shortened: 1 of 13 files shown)


bauer-andreas (dismissed):
Great catch 🚀

src/lib/util/checkName.ts
```diff
@@ -6,6 +6,11 @@
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-module-boundary-types
 export function checkName(displayName: string, name: string, value: any): void {
+  if (!value) {
```
donavanbecker: Is this log needed? @Supereg

hjdhjd (dismissed):
Only suggestion would be to ensure the verbiage is consistent with other warning/error messages in format/wording (I presume it is?).

src/lib/util/checkName.ts
donavanbecker: @hjdhjd Do you think this new log is needed?

donavanbecker:
> Only suggestion would be to ensure the verbiage is consistent with other warning/error messages in format/wording (I presume it is?).

I have removed the log and added the validation to the check.

PR #1009, opened 2023-06-27
Implement warning messages for invalid characters in names

NorthernMan54:
## :recycle: Current situation

Invalid characters in names will trigger warning messages in the Home App. #1008 

## :bulb: Proposed solution

Implement warning messages for invalid characters in names

Sample warning messages

```
HAP-NodeJS WARNING: The service 'Play (Yamaha)' is getting published with the characteristic 'Name' not following HomeKit naming rules ('Play (Yamaha)'). Use only alphanumeric, space, and apostrophe characters, start and end with an alphabetic or numeric character, and don't include emojis. This might prevent the accessory from being added to the Home App or leading to the accessory being unresponsive!
HAP-NodeJS WARNING: The service 'Pause (Yamaha)' is getting published with the characteristic 'Name' not following HomeKit naming rules ('Pause (Yamaha)'). Use only alphanumeric, space, and apostrophe characters, start and end with an alphabetic or numeric character, and don't include emojis. This might prevent the accessory from being added to the Home App or leading to the accessory being unresponsive!
HAP-NodeJS WARNING: The service 'Skip Fwd (Yamaha)' is getting published with the characteristic 'Name' not following HomeKit naming rules ('Skip Fwd (Yamaha)'). Use only alphanumeric, space, and apostrophe characters, start and end with an alphabetic or numeric character, and don't include emojis. This might prevent the accessory from being added to the Home App or leading to the accessory being unresponsive!
HAP-NodeJS WARNING: The service 'Skip Rev (Yamaha)' is getting published with the characteristic 'Name' not following HomeKit naming rules ('Skip Rev (Yamaha)'). Use only alphanumeric, space, and apostrophe characters, start and end with an alphabetic or numeric character, and don't include emojis. This might prevent the accessory from being added to the Home App or leading to the accessory being unresponsive!
HAP-NodeJS WARNING: The accessory 'Spotify (Yamaha)' is getting published with the characteristic 'name' not following HomeKit naming rules ('Spotify (Yamaha)'). Use only alphanumeric, space, and apostrophe characters, start and end with an alphabetic or numeric character, and don't include emojis. This might prevent the accessory from being added to the Home App or leading to the accessory being unresponsive!
```

## :gear: Release Notes

Warning messages will now be logged for names containing invalid characters

## :heavy_plus_sign: Additional Information
*If applicable, provide additional context in this section.*

### Testing

Testing has been augmented with test cases for the various scenarios.

Output from testing with one of my plugins that has the issue.

```
[...]

PR #1052, opened 2024-07-04
fix: ensure we check names using the full UTF-8 character set.

hjdhjd:

This further refines #1009 to:

1. Check names using the UTF-8 character set rather than ASCII.
2. Ensure we fully comply with HomeKit naming requirements.
3. Ignores the value if it’s empty. ConfiguredName is an optional characteristic and it’s okay if it’s empty, for instance.
4. Updates tsconfig to support Node 18 at a minimum (required to get UTF-8 regular expressions).

## :bulb: Proposed solution

Refactored and implemented.

## :gear: Release Notes

improvement: refinements to HomeKit name validation.

*If applicable, provide additional context in this section.*

### Testing

*Which tests were added? Which existing tests were adapted/changed? Which situations are covered, and what edge cases are missing?*

### Reviewer Nudging

@NorthernMan54

### DIFF
--- a/src/lib/util/checkName.ts
+++ b/src/lib/util/checkName.ts
@@ -1,3 +1,5 @@
+import { CharacteristicValue, Nullable } from "../../types";
+
 /**
  * Checks that supplied field meets Apple HomeKit naming rules
  * https://developer.apple.com/design/human-interface-guidelines/homekit#Help-people-choose-useful-names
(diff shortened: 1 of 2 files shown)


donavanbecker:
CheckName is used for both Name and ConfiguredName, so I feel like what is done in #1049 would be better not?

src/lib/Characteristic.ts

donavanbecker (dismissed):

PR #1054, opened 2024-07-04
fix: Update characteristic naming convention warning message text and associated tests.

hjdhjd:
This PR:

* Updates the warning message related to naming issues somewhat (a bit cleaner I think now).
* Updates all the relevant tests to ensure they match against the revised warning messages.
* Provides some additional test cleanup in spots (warnings were being emitted and not properly ignored in some of the accessory tests).
* Some minor cleanup items (assigning variables instead of just using the result) when executing checkValue et al.

### DIFF
--- a/src/lib/Accessory.spec.ts
+++ b/src/lib/Accessory.spec.ts
@@ -409,14 +409,14 @@ describe("Accessory", () => {
   });
 
   describe("Accessory and Service naming checks", () => {
-    let consoleLogSpy: jest.SpyInstance;
+    let consoleWarnSpy: jest.SpyInstance;
 
     beforeEach(() => {
-      consoleLogSpy = jest.spyOn(console, "warn");
+      consoleWarnSpy = jest.spyOn(console, "warn").mockImplementation(() => {});
     });
   
     afterEach(() => {
-      consoleLogSpy.mockRestore();
+      consoleWarnSpy.mockRestore();
     });
 
     test("Accessory Name ending with !", async () => {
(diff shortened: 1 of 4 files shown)


bwp91 (approved):

donavanbecker (commented):
Thanks!

n0rt0nthec4t:
The regex in here "/^[\p{L}\p{N}][\p{L}\p{N} ']*[\p{L}\p{N}]$/u" is too tightly limited and is excluding/warning on valid HomeKit characters in names ie: -
Try running the sample accessories such as Wi-Fi Router and Wi-Fi Satellite will complain about the naming

Suggest maybe
/^[\p{L}\p{N}][\p{L}\p{N} -,.']*[\p{L}\p{N}]$/u

-,. are still valid in HomeKit Names

hjdhjd:
> The regex in here "/^[\p{L}\p{N}][\p{L}\p{N} ']*[\p{L}\p{N}]$/u" is too tightly limited and is excluding/warning on valid HomeKit characters in names ie: - Try running the sample accessories such as Wi-Fi Router and Wi-Fi Satellite will complain about the naming
> 
> Suggest maybe /^[\p{L}\p{N}][\p{L}\p{N} -,.']*[\p{L}\p{N}]$/u
> 
> -,. are still valid in HomeKit Names

They are **not** valid characters in names. The examples should be updated. While HomeKit may accept stuff that is invalid, it’s still invalid and it is correct to warn users accordingly. 😄 [The only valid names within HomeKit, straight from the source](https://developer.apple.com/design/human-interface-guidelines/homekit#Help-people-choose-useful-names):

> Check that the names people provide follow HomeKit naming rules. If your app lets people rename services, make sure that the new names follow the rules. (The system-provided setup flow automatically checks the original names.) If people enter a name that breaks one or more rules, briefly explain the problem and suggest some alternative names that work. Here are the rules:
> * Use only alphanumeric, space, and apostrophe characters.
> * Start and end with an alphabetic or numeric character.
> * Don’t include emojis.
> 

If you can point me to more current documentation from Apple that explicitly states differently, happy to look into it, but for the time being, we’re following the north star Apple’s laid out here. Just because HomeKit will accept a character, in general, Apple sets these guidelines up for a reason…sometimes with clear reasons, often times not, but over time the reasons for those guidelines typically become clear.

Appreciate you highlighting that we have examples that need to be updated and we’ll do so.

n0rt0nthec4t:
I actually disagree that what I highlight a not valid names. While HomeKit accepts those characters in names (as it does so far in iOS 17 and iOS 18) Hapnodejs should still allow those characters. 

If we're going to be "stubborn" on this, then at least provide a mechanism for the accessory developer to disable this warning

hjdhjd:
We have no plans to do so. What you refer to as “valid names” are explicitly **not** valid. Just because HomeKit may accept them currently doesn’t make them valid.

You’re always welcome to use earlier versions of HAP-NodeJS if you prefer, but our goal is to be compliant with Apple conventions and guidelines, not merely what HomeKit will or won’t accept in a given version of the ecosystem.

PR #1055, opened 2024-07-06
Increase snapshot handler warning timeout to 8000ms.

hjdhjd:

When a camera image snapshot takes more than 5s to return an image to HAP-NodeJS a warning is generated, even though the HomeKit timeout appears to be at least 10s.
## :bulb: Proposed solution

Revised the warning to generate at 8s instead of 5s, allowing developers a longer window without warnings.

## :gear: Release Notes

Updated image snapshot warning timeout to 8s (previously 5s).

### DIFF
--- a/src/lib/controller/CameraController.ts
+++ b/src/lib/controller/CameraController.ts
@@ -1025,7 +1025,7 @@ export class CameraController extends EventEmitter implements SerializableContro
           reject(HAPStatus.OPERATION_TIMED_OUT);
         }, 17000);
         timeout.unref();
-      }, 5000);
+      }, 8000);
       timeout.unref();
 
       try {


donavanbecker (approved):

Shaquu (approved):

src/lib/controller/CameraController.ts
```diff
@@ -1025,7 +1025,7 @@ export class CameraController extends EventEmitter implements SerializableContro
           reject(HAPStatus.OPERATION_TIMED_OUT);
         }, 17000);
         timeout.unref();
-      }, 5000);
+      }, 8000);
```
bauer-andreas, line 1028: This timeout is for the „slow to respond“ message. The problem with snapshot handlers is that they will block the accessory from receiving further request from the device. 
Also currently, we do a 5000+17000 max timeout (after which we return a HAPStatue Timeout). If we really are increasing the timeout to 8s we might want to decrease the 17s one?

hjdhjd: I opted to leave the one at 17s because it seems reasonable. I understand that they may block the accessory from receiving further requests - that’s the plugin developer’s problem, not Homebridge’s. I don’t want Homebridge telling me what I already know. 😄 In my use case, I **know** that they may take longer than 5 seconds to complete - that doesn’t make it wrong or invalid and it leads to unnecessary end-user questions. There’s currently no way to turn these off or configure them to more properly be tuned to an individual plugin’s needs. That’s for another day and then the plugin developer can decide how much alerting (or not) they want. That’s the right long-term answer, but in the short-term, I think this is reasonable.

TL;DR: I think 17s is good as it is, with a warning at the ~45% mark (8s) and a timeout at 17s (100% for the purposes of this conversation). A warning at the ~17% mark (5s) is overly aggressive. It’s a reasonable compromise.

bauer-andreas: Sounds good, thanks for explaining the rationale behind this change 👍

PR #1067, opened 2024-07-21
v1.1.0

donavanbecker:
### Changes

- Set `Ciao` as the default Advertiser

### Other Changes

- Update docs
- Updated dependencies


bwp91 (approved):
We should add a very visible note to https://github.com/homebridge/homebridge/wiki/Updating-To-Homebridge-v2 about this change.

That the default advertiser has changed - and if a user wants to continue using bonjour then they should specifically set this BEFORE updating to HB2.0

donavanbecker:
> We should add a very visible note to https://github.com/homebridge/homebridge/wiki/Updating-To-Homebridge-v2 about this change.
> 
> That the default advertiser has changed - and if a user wants to continue using bonjour then they should specifically set this BEFORE updating to HB2.0

That is a good idea, I have added a note. Let me know if you think we should add more.

n0rt0nthec4t:
Seems this release locked the node engine versions to either 18.x.x or 20.x.x. Wondering what the reason for that? Latest docker node container is using 22.5.1 which prevents this from running. Should the package.json engine version be changed to >= 18.0.0?

donavanbecker:
> Seems this release locked the node engine versions to either 18.x.x or 20.x.x. Wondering what the reason for that? Latest docker node container is using 22.5.1 which prevents this from running. Should the package.json engine version be changed to >= 18.0.0?

`^18 || ^20` was set with v1.0.0. We wouldn't want to set it with `>=18` because we don't support odd node releases and node 22 isn't LTS yet so we haven't added it to the engine yet. Once 22 is LTS, then we will added it as `^18 || ^20 || ^22` or `^20 || ^22`

PR #1086, opened 2025-03-22
add constants for `SecuritySystemAlarmType`

bwp91:
I'm assuming that `0` and `1` correspond to these:

```ts
public static readonly NO_ALARM = 0;
public static readonly UNKNOWN = 1;
```


ebaauw (approved):
Yes, they do.  Thanks!

I'm afraid I fail to understand how the dashes and dots in the generator configuration translate to the camelcase characteristic name and uppercase value constants, but good to see how it's done.

PR #1087, opened 2025-03-22
update def file from `V=880` to `V=886`

bwp91:
I have been careful to avoid breaking changes but adding any removed services/characteristics into the custom lists, so that in the end, nothing is removed from the output type files.


NorthernMan54 (approved):
A lot of little changes to obscure services and characteristics.  Fingers crossed on not breaking something.

PS no vacuum cleaner.

PR #1095, opened 2025-06-13
Update Engines for NodeJS 24

NorthernMan54:


bwp91:
1. do we want to consider dropping 18
2. we should add an entry in the chanelog for this

bwp91 (approved):

